PR #36, opened 2023-03-09
refactor: JUnit Jupiter best practices

yeikel:


timtebeek:
Thanks @yeikel ! We should indeed be dog fooding more. :)
How was the experience doing this through https://public.moderne.io ?

PR #44, opened 2023-03-16
AddImport not working for Kotlin

nomisRev:
`AddImport` seems to only work for Java, and this PR adds a failing test case ported from [OpenRewrite/Rewrite - AddImportTest](https://github.com/openrewrite/rewrite/blob/main/rewrite-java-test/src/test/java/org/openrewrite/java/AddImportTest.java).

I encountered this when trying to accomplish the following in https://github.com/nomisRev/rewrite-arrow/pull/3:
 - `EffectScope` is an `interface` with a _method_ `ensure`.
 - `Raise` is an `interface` with a _top-level extension_ `ensure`.
 - After executing `ChangeType` from `EffectScope` to `Raise` the `ensure` import needs to be added.

```java
kotlin(
  """
    package com.yourorg
          
    import arrow.core.continuations.EffectScope

    suspend fun EffectScope<Int>.test(): Unit =
      ensure(false) { -1 }
    """,
  """
    package com.yourorg
          
    import arrow.core.raise.Raise
    import arrow.core.raise.ensure

    suspend fun Raise<Int>.test(): Unit =
      ensure(false) { -1 }
    """
)
```

### DIFF
--- a/src/test/java/org/openrewrite/kotlin/AddImportTest.java
+++ b/src/test/java/org/openrewrite/kotlin/AddImportTest.java
@@ -1,0 +1,32 @@
+package org.openrewrite.kotlin;
+
+import org.junit.jupiter.api.Test;
+import org.openrewrite.java.AddImport;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.kotlin.tree.ParserAssertions.kotlin;
+import static org.openrewrite.test.RewriteTest.toRecipe;
+
+public class AddImportTest  implements RewriteTest {
+    @Test
+    void addImportBeforeImportWithSameInsertIndex() {
+        rewriteRun(
+          spec -> spec.recipe(toRecipe(() -> new AddImport<>("org.junit.jupiter.api.Assertions", "assertFalse", false))),
+          kotlin(
+            """
+              import org.junit.jupiter.api.Assertions.assertTrue
+              import org.junit.Test
+
+              class MyTest
+              """,
+            """
+              import org.junit.jupiter.api.Assertions.assertFalse
+              import org.junit.jupiter.api.Assertions.assertTrue
+              import org.junit.Test
+
+              class MyTest
+              """
+          )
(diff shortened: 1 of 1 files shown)


traceyyoshima:
Hi @nomisRev, I added an AddImportTest to show that the recipe works. So that you know, I had AddImport working on rewrite-arrow with your targeted imports.

nomisRev:
Okay, got it @traceyyoshima! Thank you.

I was very confused on how to add imports for _top-level function_ since on Java they're considered "static", but you can just put the _top-level import_ in `fullyQualifiedName`, and use `null` for `statik`.

In my case, it seems I have to rely on `onlyIfReferenced = false`, or I think I'd need to write a custom recipe that combines `ChangeType` with `maybeAddImport`? The import is only needed *after* the `ChangeType` rewrite.

traceyyoshima:
Hi @nomisRev, I pushed some changes to the branch that will work.

For now, I adjusted the conditions to work on java, but we may need to add an AddImport for `rewrite-kotlin`. To your point, the java AddImport doesn't detect syntax like `arrow.core.raise.ensure`, so it fails to find types in use.

I think you're right, it may be best to add an imperative recipe that contains all the conditions to asses safe changes, what changes are necessary, and call the appropriate recipes when necessary.

nomisRev:
Thank you so much for your help @traceyyoshima 🙌  This has been great ☺️ I made a ton of progress this week, and confident I almost got everything.

I've also managed to get the _change top-level method_ working based on your suggestions. Can I tag your for a review if I think I have it somewhat working like I think it should? 😬

traceyyoshima:
> I've also managed to get the _change top-level method_ working based on your suggestions. Can I tag your for a review if I think I have it somewhat working like I think it should? 😬

Hi @nomisRev, you're welcome :), happy to help and I'm glad you've made good progress.

Very cool, yes, I'd be glad to take a look.

PR #50, opened 2023-03-17
Added ChangeTypeAlias recipe.

### LINKED ISSUE #48: ChangeType not working for typealias
I am using `ChangeType` to rewrite `Validated` to `Either`, and it's not working for following case. (It's perhaps a bit far-fetched and wasn't expecting this to work, but implementing a custom `ChangeType` seems tricky so hoping we can perhaps find an intermediate solution).

I currently have following typealiases:

```kotlin
typealias ValidatedNel<E, A> = Validated<NonEmptyList<A>, E>
typealias EitherNel<E, A> = Either<NonEmptyList<A>, E>
```

And following setting in my `yml`:

```yml
- org.openrewrite.java.ChangeType:
    oldFullyQualifiedTypeName: arrow.core.Validated
    newFullyQualifiedTypeName: arrow.core.Either
```

I was hoping it would rewrite:

```kotlin
val x: ValidatedNel<String, Int> = TODO()
```

to

```kotlin
val x: Either<NonEmptyList<String>, Int> = TODO()
```

Ideally would be a way to rewrite from `ValidatedNel<E, A>` to `EitherNel<E, A>` but it's not at all blocking.
I also tried following, but naturally it didn't do anything 😁 

```yml
- org.openrewrite.java.ChangeType:
    oldFullyQualifiedTypeName: arrow.core.ValidatedNel
    newFullyQualifiedTypeName: arrow.core.Either
```

traceyyoshima:
fixes #48

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/ChangeTypeAlias.java
+++ b/src/main/java/org/openrewrite/kotlin/ChangeTypeAlias.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.openrewrite.kotlin;
 
 import lombok.EqualsAndHashCode;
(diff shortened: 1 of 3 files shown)


traceyyoshima:
Quick draft of renaming aliases ... TypeAliases are an odd change because the Alias Type does not exist on other identifiers. It may require naming utilities and RenameVariable only to rename applicable references.

traceyyoshima:
@nomisRev, what do you think about changing an alias based on the expected type?

ChangeTypeAlias API:
- aliasName
- newName
- fullyQualifiedAliasedType

The compiler does not attribute a new type for a `typealias`, so the type is based on the alias's assigned type. I.E. in `typealias TestAlias<T> = Test<T>` `TestAlias` will have a type of `Test`. and the type expression of `val a: TestAlias<String> = Test<String>()` `TestAlias` will have a type of `Test.

However, we can detect when an alias is declared and check if the type matches our target FQN. Then use RenameVariable to ensure all identifiers in scope are renamed.

Note: I'm not familiar with `typealias` and this recipe assumes the alias is limited to a local scope in the SourceFile.

nomisRev:
> Note: I'm not familiar with typealias and this recipe assumes the alias is limited to a local scope in the SourceFile.

I think that would be too limiting, the typical use-case being that a dependency declares a `typealias` and then it needs to be rewritten to a different one?

Replacing the typealias with the actual type I think is fine.

# Arrow Use Case

Hmm, I am afraid my use-case is much more complex than this one 😓 Since not only the `typealias` changes, but also the underlying type. The `ChangeType` is also involved, but I guess I need to also make a compound recipe for this?

```kotlin
typealias ValidatedNel<E, A> = Validated<NonEmptyList<E>, A>
```
to 
```kotlin
typealias EitherNel<E, A> = Either<NonEmptyList<E>, A>
```

I think it would great to just support `ChangeType` for `typealias`, but it seems not to be working. I.e. if `ValidatedNel<E, A>` would rewrite to `Either<NonEmptyList<E>, A>` but then it would also have to add the appropriate import for `NonEmptyList`. Not sure if that is easier to achieve for my use-case.

It's a bit silly, but a find-and-replace would also work in my case if all other rewrites go through 😅

nomisRev:
A typealias can also appear in al places where `ChangeType` can appear, so this can be in function parameters / return, class constructors, variables, etc.

PR #52, opened 2023-03-20
Add change type alias

### LINKED ISSUE #48: ChangeType not working for typealias

traceyyoshima:
fixes #48

Note: Imports will note be changed until https://github.com/openrewrite/rewrite-kotlin/issues/53 is resolved.

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/ChangeTypeAlias.java
+++ b/src/main/java/org/openrewrite/kotlin/ChangeTypeAlias.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


traceyyoshima:
@nomisRev, I re-opened `ChangeTypeAlias` along with more tests to show that the recipe handles every case aside from Import aliases.

I've created a new issue to [discover how we may add type attribution to imports](https://github.com/openrewrite/rewrite-kotlin/issues/53), since the information does not exist on the FIR.

Note: That `ChangeTypeAlias` does a lot more than a search and replace, because it tracks name space conflicts, guarantees the name is unique, and will update all the references to the type alias while ignoring shadowed names.

PR #57, opened 2023-03-30
Add failing test destruct

nomisRev:
Destruct gets mangeld upon rewrite:

```kotlin
fun example() {
  val (a, b, c) = Triple(1, 2, 3)
}
```

gets rewritten to

```kotlin
fun example() {
  val <destruct>abc(a, b, c) = Triple(1, 2, 3)
}
```


src/test/java/org/openrewrite/kotlin/DestructuringTest.java
```diff
@@ -0,0 +1,48 @@
...
+package org.openrewrite.kotlin;
+
+import org.junit.jupiter.api.Test;
+import org.openrewrite.java.ChangeType;
+import org.openrewrite.test.RecipeSpec;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.kotlin.Assertions.kotlin;
+
+public class DestructuringTest implements RewriteTest {
```
traceyyoshima, line 25: @nomisRev, thanks for reporting the issue! This was a known issue, but I haven't had time to fix it. I closed the open issue in favor of what you reported -- similar to issue 56 would you please add the test to [VariableDeclarationsTest](https://github.com/openrewrite/rewrite-kotlin/blob/main/src/test/java/org/openrewrite/kotlin/tree/VariableDeclarationTest.java)?

Issue #60, opened 2023-04-03
Suspend lambda incorrectly rewrites return type


nomisRev:
When using the _special_ syntax for `suspend` lambdas without arguments `suspend { }` , the return type is incorrectly changed.

```kotlin
fun method ( ) {
    val lambda: suspend ( ) -> Int = suspend { 1 }
}
```

Results in:

```text
expected: 
  "fun method ( ) {
      val lambda: suspend ( ) -> Int = suspend { 1 }
  }"
 but was: 
  "fun method ( ) {
      val lambda:  Int = suspend { 1 }
  }"
```

traceyyoshima:
fixed by https://github.com/openrewrite/rewrite-kotlin/commit/7c4cdcb3350bab8bdac599c7fdcdbebd43c9a88f

Issue #56, opened 2023-03-30
Suspend lambda gets mangled during parsing/rewrite


nomisRev:
A Kotlin `suspend` lambda gets rewritten incorrectly. See https://github.com/openrewrite/rewrite-kotlin/pull/55 with failing test.

```kotlin
suspend fun example(
  title: String,
  verifyUnique: suspend (String) -> Boolean
): String = TODO()
```

becomes

```kotlin
suspend fun example(
  title: String,
  verifyUnique: Stringsuspend (String) -> Boolean
): String = TODO()
```

PR #61, opened 2023-04-03
Add tests for issue 56 and 60

nomisRev:
This PR adds two failing test cases discovered after discussion with @traceyyoshima in https://github.com/openrewrite/rewrite-kotlin/pull/55.

Relates to #56 and #60

### DIFF
--- a/src/test/java/org/openrewrite/kotlin/tree/LambdaTest.java
+++ b/src/test/java/org/openrewrite/kotlin/tree/LambdaTest.java
@@ -17,6 +17,7 @@ package org.openrewrite.kotlin.tree;
 
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
+import org.junitpioneer.jupiter.ExpectedToFail;
 import org.openrewrite.Issue;
 import org.openrewrite.test.RewriteTest;
 
@@ -72,7 +73,7 @@ class LambdaTest implements RewriteTest {
 
     @Issue("https://github.com/openrewrite/rewrite-kotlin/issues/60")
     @Test
-    @Disabled
+    @ExpectedToFail
     void suspendLambda() {
         rewriteRun(
           kotlin(
@@ -87,7 +88,7 @@ class LambdaTest implements RewriteTest {
 
     @Issue("https://github.com/openrewrite/rewrite-kotlin/issues/56")
     @Test
-    @Disabled
+    @ExpectedToFail
     void suspendLambdaWithParameter() {
         rewriteRun(
           kotlin(
--- a/src/test/java/org/openrewrite/kotlin/tree/LambdaTest.java
+++ b/src/test/java/org/openrewrite/kotlin/tree/LambdaTest.java
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.kotlin.tree;
 
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.openrewrite.Issue;
 import org.openrewrite.test.RewriteTest;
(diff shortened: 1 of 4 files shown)


src/test/java/org/openrewrite/kotlin/tree/LambdaTest.java
```diff
@@ -67,4 +69,34 @@ fun inputValues ( ) : List < Pair < String , Any ? > > {
           )
         );
     }
+
+    @Issue("https://github.com/openrewrite/rewrite-kotlin/issues/60")
+    @Test
+    @Disabled
```
timtebeek, line 75: We could adopt `@ExpectedToFail` here from junit-pioneer; that way the test will always run, and fail only when the issue is fixed, to signify that we can update and close the associated issue. Optional, but potentially nice to have.

nomisRev: Sounds like a good idea! Shall I do the same for test cases added in https://github.com/openrewrite/rewrite-kotlin/pull/55 and https://github.com/openrewrite/rewrite-kotlin/pull/57?

They're a bit unrelated to this PR.. I can also split it if preferred ☺️

timtebeek: Fine to do that here & Thanks for the quick response!

nomisRev:
@timtebeek I updated the PR, but it required me to introduce a new dependency. The repo uses `release.latest` but that wasn't working for `junit-pioneer` so I had to specify `2.0.0` explicitly.

src/test/java/org/openrewrite/kotlin/tree/LambdaTest.java
```diff
@@ -15,7 +15,10 @@
  */
 package org.openrewrite.kotlin.tree;
 
+import org.junit.jupiter.api.Disabled;
```
timtebeek, line 18:
```suggestion
```

src/test/java/org/openrewrite/kotlin/tree/MethodDeclarationTest.java
```diff
@@ -17,6 +17,7 @@
 
 import org.junit.jupiter.api.Disabled;
```
timtebeek, line 18:
```suggestion
```

src/test/java/org/openrewrite/kotlin/tree/VariableDeclarationTest.java
```diff
@@ -17,6 +17,7 @@
 
 import org.junit.jupiter.api.Disabled;
```
timtebeek, line 18:
```suggestion
```

timtebeek (approved):
I _think_ we can clean up these imports, but haven't checked; let's try & merge if this works.

nomisRev:
Oops, missed cleaning up the imports indeed. Normally I use `cmd+opt+l` in IntelliJ but that seems to format to a completely different style than I have.

Do you have any `.editorconfig` configured in OpenRewrite projects? That could make it easier for contributors to stick to the same style without changing IntelliJ configurations.

timtebeek:
> Oops, missed cleaning up the imports indeed. Normally I use `cmd+opt+l` in IntelliJ but that seems to format to a completely different style than I have.

No worries; we're working on automating these changes to lighten the review load. :)

> Do you have any `.editorconfig` configured in OpenRewrite projects? That could make it easier for contributors to stick to the same style without changing IntelliJ configurations.

We often do; don't know if that's in rewrite-kotlin already, but we have a few scattered around and plans to merge this towards openrewrite/rewrite eventually.

PR #63, opened 2023-04-04
MethodMatcher seems to have broken in 0.3.0

nomisRev:
When trying to bump to `0.3.0` I encountered 3 issues:

 - Binary breaking change in `UsesMethod`. I guess this is non-problematic for "experimental" library such as rewrite-kotlin, but was unexpected without a major bump in rewrite version.
 - BOM is outdated, but this was easily fixed by explicitly relying on versions
 - Regression in MethodMatcher, I suspect only for rewrite-kotlin

I am not 100% sure, but I think this test reproduces the problem.

### DIFF
--- a/src/test/java/org/openrewrite/kotlin/MethodMatcherTest.java
+++ b/src/test/java/org/openrewrite/kotlin/MethodMatcherTest.java
@@ -1,0 +1,25 @@
+package org.openrewrite.kotlin;
+
+import org.junit.jupiter.api.Test;
+import org.openrewrite.java.search.FindMethods;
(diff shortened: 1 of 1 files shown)


nomisRev:
Okay, I found the issue but I cannot explain how it broke. I guess it makes sense though.
I was matching the original code arrow.core.continuations.Effect fold(..) but during the yml it's already rewritten due to ChangeType(arrow.core.continuations.Effect, arrow.core.raise.Effect).
So I had to update my UsesType and MethodMatcher accordingly. So I'd guess the original behavior was actually relying on a bug? :thinking_face:

PR #102, opened 2023-04-24
Ensure `J.FieldAccess#name` has field type

knutwannheden:
Adds `J.Identifier#fieldType` type attribution for `J.FieldAccess#name` instances.

### DIFF
--- a/src/test/java/org/openrewrite/kotlin/tree/FieldAccessTest.java
+++ b/src/test/java/org/openrewrite/kotlin/tree/FieldAccessTest.java
@@ -27,7 +27,6 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.openrewrite.kotlin.tree.ParserAssertions.kotlin;
 import static org.openrewrite.test.RewriteTest.toRecipe;
 
-
 class FieldAccessTest implements RewriteTest {
 
     @Test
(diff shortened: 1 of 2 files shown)


src/test/java/org/openrewrite/kotlin/tree/FieldAccessTest.java
```diff
@@ -114,4 +121,28 @@ void qualifier() {
...
+                      assertThat(fieldType.getType().toString()).isEqualTo("kotlin.Int");
+                      assertThat(fieldType.getOwner().toString()).isEqualTo("kotlin.Int$Companion");
+                  }
+                  return super.visitFieldAccess(fieldAccess, ctx);
+              }
+          })),
+          kotlin(
+            """
+            val i = Int.MIN_VALUE
+            """
```
jkschneider, line 144: Indentation nit: indentation of end delimiter of text block should be 2 spaces to the right of start delimiter and code should be left-aligned to that end delimiter.

traceyyoshima: Meaning something like this?
```
"""
  Code here.
  """
```

If so, I'll sweep through when I have time to fix the tests -- the current PR matches the rest of the code

knutwannheden:
@traceyyoshima I rebased it on my other PR which was merged.

src/test/java/org/openrewrite/kotlin/tree/FieldAccessTest.java
```diff
@@ -114,4 +121,28 @@ void qualifier() {
...
+          )
+        );
+    }
```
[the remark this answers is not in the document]
traceyyoshima: Generally, type mapping tests work through the KotlinTypeGoat, since we can check the type signatures generated in the KotlinTypeSignatureBuilderTest.

I think this is fine for now, I'll add it to KotlinTypeGoat when I have time.

traceyyoshima (approved):
LGTM -- ty!

knutwannheden:
I will redeploy tomorrow and check if `ShortenFullyQualifiedTypeReferences` now behaves correctly for Kotlin.

I suspect that there are other recipes which might now also behave differently for Kotlin, now that there is more type attribution.

PR #103, opened 2023-04-24
Basic support for platform-typed objects

knutwannheden:
So that the type which is derived from a platform type doesn't end up as `Unknown`, the corresponding lower bound is used instead (for a platform type `String!` the lower bound is `String` and the upper bound is `String?`).


traceyyoshima (approved):
Looks good -- ty!

Issue #105, opened 2023-04-28
Java annotation problems with rewrite-kotlin


fpoyer:
Context: working on a huge java/kotlin legacy code base, and trying to apply the PowerMockitoToMocktio recipe (as a starter, but I expect to do a whole lot of things after that, if I can get open rewrite to work).

Problem(s): about 50% of my Kotlin files are not parsed, and when they are, not always refactored as expected.

I cloned the repo and add the following tests to the AnnotationTest test class, all fails except the last one, somehow...:
```java
    @Test
    void fullyQualifiedCustomAnnotation() {
        rewriteRun(
          java(
            """
              package org.powermock.core.classloader.annotations;
                            
              import java.lang.annotation.Target;
              import static java.lang.annotation.ElementType.*;
                            
              @Target({TYPE, FIELD, METHOD, PARAMETER, CONSTRUCTOR, LOCAL_VARIABLE})
              public @interface PowerMockIgnore {
              }
              """
          ),
          // Works
          kotlin(
            """
              import org.powermock.core.classloader.annotations.PowerMockIgnore

              @PowerMockIgnore
              class A {
              }
              """),
          // Fails, despite being valid Kotlin code
          kotlin(
            """
              @org.powermock.core.classloader.annotations.PowerMockIgnore
              class A {
              }
              """)
        );
    }

    @Test
    void fullyQualifieJavaLangdAnnotation() {
        rewriteRun(
          // Works, Deprecated is defined in java.lang package so no import needed
          kotlin(
            """
              @Deprecated
              class A {
              }
              """),
          // Fails, despite being valid Kotlin code
          kotlin(
            """
              @java.lang.Deprecated
              class A {
              }
              """)
        );
    }

    // This one fails, but I guess this is because of the missing imports (coming from external libraries that are not on the runtime classpath)
    @Test
    void doesNotParseAndPrintWeirdStuffAfterClass() {
        rewriteRun(
          kotlin(
            """
              @Api(value = "some request"
              @Path("/v1/endpoint")
              @Produces(MediaType.APPLICATION_JSON)
              @Consumes(MediaType.APPLICATION_JSON)
              @RestController
              class A {
              }
                 """
          )
        );
    }

    // Despite being recognized as an annotation, it is not removed (but it works fine with annotations from java.lang package)
    @Test
    void customAnnotationIsNotRemoved() {
        rewriteRun(
[...]

PR #106, opened 2023-05-02
Add tests that replicate annotation parsing and printing problems

fpoyer:
Context: #105 Problems, most likely caused by parsing errors of the Kotlin code, while trying to apply the PowerMockitoToMockito recipe
If some of the problems appear to be caused by the recipe itself, I'll open an issue in the `rewrite-testing-frameworks` repository.

Feel free to edit, annotate, comment, etc


src/test/java/org/openrewrite/kotlin/tree/AnnotationTest.java
```diff
@@ -82,4 +84,123 @@ void jvmNameAnnotation() {
...
+            """
+              import com.example.Test
+                          
+              @Test( values = [ "a" , "b" , "c" ] )
+              val a = 42
+              """,
+            """
+              
+              
+
```
timtebeek, line 200: Perhaps we'd rather see this fail right? As the excess new lines aren't desirable in the output.
```suggestion
```

src/test/java/org/openrewrite/kotlin/tree/AnnotationTest.java
```diff
@@ -82,4 +84,123 @@ void jvmNameAnnotation() {
...
+              """
+            , """
+              class A {
+              }
+              """)
+        );
+    }
+
+    // Somehow, this one works? Maybe because of the parameters?
+    // Mind the required empty lines after refactor though...
```
timtebeek, line 175:
```suggestion
```

src/test/java/org/openrewrite/kotlin/tree/AnnotationTest.java
```diff
@@ -82,4 +84,123 @@ void jvmNameAnnotation() {
           )
         );
     }
+
+    @Test
```
timtebeek, line 88:
```suggestion
    @Test
    @Issue("https://github.com/openrewrite/rewrite-kotlin/issues/105")
```

src/test/java/org/openrewrite/kotlin/tree/AnnotationTest.java
```diff
@@ -82,4 +84,123 @@ void jvmNameAnnotation() {
...
+          kotlin(
+            """
+              @com.example.annotations.JavaAnnotationToRemove
+              class A {
+              }
+              """)
+        );
+    }
+
+    @Test
```
timtebeek, line 122:
```suggestion
    @Test
    @Issue("https://github.com/openrewrite/rewrite-kotlin/issues/105")
```

src/test/java/org/openrewrite/kotlin/tree/AnnotationTest.java
```diff
@@ -82,4 +84,123 @@ void jvmNameAnnotation() {
...
+            """
+              @java.lang.Deprecated
+              class A {
+              }
+              """)
+        );
+    }
+
+    // Despite being recognized as an annotation, it is not removed (but it works fine with annotations from java.lang package)
+    @Test
```
timtebeek, line 143:
```suggestion
    @Test
    @Issue("https://github.com/openrewrite/rewrite-kotlin/issues/105")
```

src/test/java/org/openrewrite/kotlin/tree/AnnotationTest.java
```diff
@@ -82,4 +84,123 @@ void jvmNameAnnotation() {
...
+            , """
+              class A {
+              }
+              """)
+        );
+    }
+
+    // Somehow, this one works? Maybe because of the parameters?
+    // Mind the required empty lines after refactor though...
+    @Test
```
timtebeek, line 176:
```suggestion
    @Test
    @Issue("https://github.com/openrewrite/rewrite-kotlin/issues/105")
```

src/test/java/org/openrewrite/kotlin/tree/AnnotationTest.java
```diff
@@ -82,4 +84,123 @@ void jvmNameAnnotation() {
...
+          kotlin(
+            """
+              @java.lang.Deprecated
+              class A {
+              }
+              """)
+        );
+    }
+
+    // Despite being recognized as an annotation, it is not removed (but it works fine with annotations from java.lang package)
```
timtebeek, line 142: Appreciate the clarification here while we work through these failures; typically we put that context in the issue before we merge, and then have the test method name reflect what we (don't) want to happen.

timtebeek (approved):
Some quick suggestions, and OK'ed the tests to run. Thanks for adding these tests already!

timtebeek:
@traceyyoshima What flow would you prefer to work through these issues? @fpoyer has offered to help as detailed in #105 and https://github.com/openrewrite/rewrite/discussions/3173 . We can tag these tests as [`@ExpectedToFail`](https://junit-pioneer.org/docs/expected-to-fail-tests/) and merge early, or try to resolve some of these before we actually merge.

fpoyer:
Tried walking with debugger to see if I can understand what is wrong.
For the "fully qualified annotation" problem, I think problem is in the [KotlinParserVisitor](https://github.com/openrewrite/rewrite-kotlin/blob/fb963ccdd2ffd420b7df11f37c40476547055d80/src/main/java/org/openrewrite/kotlin/internal/KotlinParserVisitor.java#L171), where the annotation itself if treated as an "ErrorNamedReference" (dunno if this is expected or not) whose fullname is "JavaAnnotationToRemove" instead of "com.example.annotations.JavaAnnotationToRemove", and thus the following "skip" doesn't work as expected.

I guess this will be the same for the fully qualified java.lang annotation.

But I have no idea how to fix that...

timtebeek:
Had a brief look at `org.openrewrite.kotlin.tree.AnnotationTest#fullyQualifieJavaLangAnnotation` in the debugger here as well; Interesting thing is that there's a clear difference between the two cases:

### `@Deprecated`
Notice how the name here is lacking a type.
Screenshot from 2023-05-03 10-36-55

### `@java.lang.Deprecated`
Whereas the name is does have a type.
image

Then when it comes to printing the source code out again, we see that result in `@Deprecatedjava.lang.Deprecated` for the typed case:
```
When parsing and printing the source code back to text without modifications, the printed source didn't match the original source code. This means there is a bug in the parser implementation itself. Please open an issue to report this, providing a sample of the code that generated this error!] 
expected: 
  "@java.lang.Deprecated
  class A {
  }"
 but was: 
  "@Deprecatedjava.lang.Deprecated
  class A {
  }"
```

So there might be an issue with printing back out typed annotations, as a start, which might help resolve that particular case.

src/test/java/org/openrewrite/kotlin/tree/AnnotationTest.java
```diff
@@ -82,4 +84,122 @@ void jvmNameAnnotation() {
...
+    }
+
+    @Test
+    @Issue("https://github.com/openrewrite/rewrite-kotlin/issues/105")
+    void fullyQualifieJavaLangAnnotation() {
+        rewriteRun(
+          // Works, Deprecated is defined in java.lang package so no import needed
+          kotlin(
+            """
+              @Deprecated
```
timtebeek, line 130: In my IDE this resolves to `kotlin.Deprecated` (missing the required message argument), not `java.lang.Deprecated`. Only when I add `import java.lang.Deprecated` does this resolve to the Java type.

We likely should update the test to either add that import or required message argument.

src/test/java/org/openrewrite/kotlin/tree/AnnotationTest.java
```diff
@@ -82,4 +84,122 @@ void jvmNameAnnotation() {
...
+              }
+              """)
+        );
+    }
+
+    @Test
+    @Issue("https://github.com/openrewrite/rewrite-kotlin/issues/105")
+    void fullyQualifieJavaLangAnnotation() {
+        rewriteRun(
+          // Works, Deprecated is defined in java.lang package so no import needed
```
timtebeek, line 127: See below; comment is slightly misleading about which deprecated annotation is picked up.

timtebeek:
@traceyyoshima I can see the `java.lang.Deprecated` appended to `@Deprecated` when I place a breakpoint on this exact line https://github.com/openrewrite/rewrite-kotlin/blob/fb963ccdd2ffd420b7df11f37c40476547055d80/src/main/java/org/openrewrite/kotlin/internal/KotlinPrinter.java#L350
Any idea what might be going on there? Really counter intuitive to see that on a line printing "space", so I might just be missing background.

fpoyer:
Added some tests in main project (see https://github.com/openrewrite/rewrite/pull/3195) to make sure FQNs are correctly handled (they are), thus restricting the problem to Kotlin parsing, hope that helps.

traceyyoshima:
> @traceyyoshima What flow would you prefer to work through these issues? @fpoyer has offered to help as detailed in #105 and [openrewrite/rewrite#3173](https://github.com/openrewrite/rewrite/discussions/3173) . We can tag these tests as [`@ExpectedToFail`](https://junit-pioneer.org/docs/expected-to-fail-tests/) and merge early, or try to resolve some of these before we actually merge.

Adding `@ExpectedToFail` along with an `@Issue(...)` and merging early would be great! I'm focused on TypeScript at the moment and will switch to Kotlin/Python afterward.

traceyyoshima:
> Any idea what might be going on there? Really counter intuitive to see that on a line printing "space", so I might just be missing background.

Based on what I've read, I suspect there are a few issues:
1. The `J.ClassDeclaration.Kind` has annotations too. The annotations on class Kind are probably already printed where the breakpoint is set.
2. Parsing / printing `J.Annotation` needs a slight adjustment.
3. The lack of changes is caused by missing type attribution because JavaTypes haven't been mapped yet in [KotlinTypeMapping](https://github.com/openrewrite/rewrite-kotlin/blob/main/src/main/java/org/openrewrite/kotlin/KotlinTypeMapping.java#L119).

Mapping the `Java` types is very complicated due to differences in how `Kotlin` represents the types, so it hasn't been prioritized until it's necessary. I think it would be helpful to have an issue `Add TypeMapping for Java types` with reference to https://github.com/openrewrite/rewrite-kotlin/issues/105.

The tests won't pass until `FirJavaTypeRef` from the `Kotlin` compiler is mapped to a `JavaType`.

timtebeek (approved):
The current failures are also on the main branch, so this one is good to merge I think.

PR #125, opened 2023-06-19
Remove duplicate static analysis recipe

mike-solomon:
The other one exists in the rewrite-static-analysis repo: https://github.com/openrewrite/rewrite-static-analysis/blob/main/src/main/resources/META-INF/rewrite/common-static-analysis.yml#L96


sambsnyd:
Thanks @mike-solomon ! Appreciate you noticing and cleaning that up

PR #134, opened 2023-06-26
fix the checkNotNull `!!` missing 

### LINKED ISSUE #135: check-not-null operator `!!` missing when printing
check-not-null operator `!!` missing when printing

Example case:
```kotlin
fun foo() {
    val l = listOf ( "x" )
    val a = l [ 0 ] !!
}
```

kunli2:
Fixes #135 
1. Fix the checkNotNull `!!` missing 
2. Add a failing test of `extensionFunction`

### DIFF
--- a/src/test/java/org/openrewrite/kotlin/tree/ExtensionFunctionTest.java
+++ b/src/test/java/org/openrewrite/kotlin/tree/ExtensionFunctionTest.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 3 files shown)


src/test/java/org/openrewrite/kotlin/tree/ExtensionFunctionTest.java
```diff
@@ -0,0 +1,37 @@
...
+
+class ExtensionFunctionTest implements RewriteTest {
+
+    @Disabled
+    @Test
+    void extensionFunction() {
+        rewriteRun(
+          kotlin(
+            """
+              fun String.escape() = this.replace("a", "b")"
```
traceyyoshima, line 32:
```suggestion
              fun String . escape ( ) = this . replace ( "a" , "b" )"
```

traceyyoshima: We typically add spaces where possible in the `tree` tests to assert the whitespace is preserved

src/test/java/org/openrewrite/kotlin/tree/VariableDeclarationTest.java
```diff
@@ -370,4 +370,17 @@ class Test {
...
         );
     }
+
+    @Test
+    void nonNull() {
+        rewriteRun(
+          kotlin("""
+            fun foo() {
+                val l = listOf("x")
+                val a = l[0] !!
```
traceyyoshima, line 380:
```suggestion
                val a = l [ 0 ] !!
```

src/test/java/org/openrewrite/kotlin/tree/ExtensionFunctionTest.java
kunli2: I see.

traceyyoshima (approved):

Issue #129, opened 2023-06-23
Extension function parsing error


kunli2:
This is a failing test 

```java
    @Test
    void extensionFunction() {
        rewriteRun(
          kotlin(
            """
              fun String.escape() = this.replace("a", "b")"
              """
          )
        );
    }
```

Error:
Expected :"fun String.escape() = this.replace("a", "b")""
Actual   :"fun String.escape() = replacethis.replace("a"), "b")""

kunli2:
if remove `this` keyword, then got another parsing error

PR #142, opened 2023-06-27
Fix ExtensionFunction parsing error

kunli2:
fixes https://github.com/openrewrite/rewrite-kotlin/issues/129 

This PR is to fix `ExtensionFunction` with `this` as an explicit receiver parsing error.

Here is an example:
```kotlin
fun String . escape ( ) = this . replace ( "a" , "b" )
```

Error is:

Expected no change, but got 
```diff
---fun String . escape ( ) = this . replace ( "a" , "b" )
+++fun String . escape ( ) = replacethis . replace ( "a") , "b" )
```


traceyyoshima (approved):
LGTM, nice work!

PR #154, opened 2023-07-03
Revised `KotlinParserVisitor` to access PSI.

### LINKED ISSUE #138: Parsing error on when expressions surrounded by parens.
Note: The FIR removes unnecessary parens.
```java
    @Test
    void inParens() {
        rewriteRun(
          kotlin(
            """
              fun method(a: Any) {
                   val any = (if (a is Boolean) "true" else "false")
               }
              """
          )
        );
    }
```

### LINKED ISSUE #143: Detect FIR unary prefix and postfix operators and unwrap the FirBlock.
The FIR wraps unary operations in a FirBlock and adds extra statements.

When a unary operator is used in a statement that expects an expression, the FirBlock is converted to a J.Block. Kotlin allows statements in places Java only allows Expressions, so the Statement is wrapped in a `K.StatementExpression.` The result is a `K.StatementExpression` that contains a `J.Block` with a single statement of a `J.Unary.`

The parser should detect when unary operators are wrapped in an FIR block and return the correct expression instead.

traceyyoshima:
Changes:

- KotlinParserVisitor contains a map of cursor positions to the Kt PSI AST nodes.
- Updated visitElement to detect Parenthesized elements and Unary operators.

fixes #138
fixes #143

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/internal/KotlinParserVisitor.java
+++ b/src/main/java/org/openrewrite/kotlin/internal/KotlinParserVisitor.java
@@ -15,10 +15,11 @@
  */
 package org.openrewrite.kotlin.internal;
 
-import org.jetbrains.kotlin.KtFakeSourceElementKind;
-import org.jetbrains.kotlin.KtLightSourceElement;
-import org.jetbrains.kotlin.KtRealPsiSourceElement;
-import org.jetbrains.kotlin.KtSourceElement;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.kotlin.*;
+import org.jetbrains.kotlin.com.intellij.lang.ASTNode;
+import org.jetbrains.kotlin.com.intellij.psi.PsiElement;
+import org.jetbrains.kotlin.com.intellij.psi.PsiElementVisitor;
 import org.jetbrains.kotlin.descriptors.ClassKind;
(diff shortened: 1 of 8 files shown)


src/main/java/org/openrewrite/kotlin/internal/KotlinParserVisitor.java
```diff
@@ -86,22 +87,47 @@ public class KotlinParserVisitor extends FirDefaultVisitor<J, ExecutionContext>
...
+                    nodes.put(element.getTextRange().getStartOffset(), element.getNode());
+                    assert source.substring(element.getTextRange().getStartOffset(), element.getTextRange().getEndOffset()).equals(element.getText());
+                }
+
+                for (PsiElement child : element.getChildren()) {
+                    if (child != null) {
+                        visitElement(child);
+                    }
+                }
+                if (element.getNextSibling() != null) {
```
kunli2, line 125: I think line 125-127 is not necessary and redundant for the current solution, since `NextSibling` is a child of the parent element visited above. Removing these can improve the performance of visiting, otherwise the visiting for a large file will be heavy

kunli2 (approved):

src/main/java/org/openrewrite/kotlin/internal/KotlinParserVisitor.java
traceyyoshima: Nice catch -- updated to prevent duplicate visits.

PR #167, opened 2023-07-10
[WIP] New mechanism to support annotation use-site pasing driven by PSI

kunli2:
As discussion, this is a new mechanism to support annotation use-site parsing by leveraging PSI.


kunli2:
discussed with Tracey offline, and we don't want to apply this mechanism for now.

PR #171, opened 2023-07-11
Fix broken for-loop parsing after we fix annotation parsing

kunli2:
https://github.com/openrewrite/rewrite-kotlin/pull/170 breaks the for-loop parsing.
This PR attempts to fix broken for-loop parsing.

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/internal/KotlinParserVisitor.java
+++ b/src/main/java/org/openrewrite/kotlin/internal/KotlinParserVisitor.java
@@ -39,6 +39,7 @@ import org.jetbrains.kotlin.fir.types.*;
 import org.jetbrains.kotlin.fir.types.impl.FirImplicitNullableAnyTypeRef;
 import org.jetbrains.kotlin.fir.types.impl.FirImplicitUnitTypeRef;
 import org.jetbrains.kotlin.fir.visitors.FirDefaultVisitor;
+import org.jetbrains.kotlin.lexer.KtKeywordToken;
 import org.jetbrains.kotlin.lexer.KtModifierKeywordToken;
 import org.jetbrains.kotlin.psi.*;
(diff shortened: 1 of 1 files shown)


src/main/java/org/openrewrite/kotlin/internal/KotlinParserVisitor.java
```diff
@@ -1621,13 +1622,45 @@ public J visitNamedArgumentExpression(FirNamedArgumentExpression namedArgumentEx
                 typeMapping.type(namedArgumentExpression.getTypeRef()));
     }
 
+    /**
+     * Check whether the giving property (include destructing) AST Node has `val` or `var` modifier.
+     * @param propertyNode the property AST Node to check
+     * @return `val` or `var` if the property node has `val` or `var` modifier. otherwise return null.
+     */
+    @Nullable
+    private static String getValOrVarModifier(@Nullable PsiElement propertyNode) {
```
traceyyoshima, line 1631: What do you think about something like this instead?

Notes:
- `val` and `var` will exist on property declarations.
- `val` and `var` are always converted to `J.Annotation`.
- Our conventions add the open brace inline at the left-hand side.
- Exit once the name token is found rather than iterate through all of the property's child elements.

```
@Nullable
private J.Annotation getPropertyModifier(@Nullable PsiElement node) {
    Iterator<PsiElement> it = PsiUtilsKt.getAllChildren(getCurrentPsiNode()).iterator();
    while (it.hasNext()) {
        PsiElement child = it.next();
        if (child instanceof LeafPsiElement) {
            if (child.getNode().getElementType() instanceof KtKeywordToken &&
                    child.getText().equals("val") || child.getText().equals("var")) {
                return mapKModifierToAnnotation(child.getText());
            } else if (child.getText().equals(property.getName().asString())) {
                break;
            }
        }
    }
    return null;
}
```

Issue #173, opened 2023-07-11
Parsing error of constructor parameters with annotation


kunli2:
Parsing of constructor parameters with annotation is not supported.

Example:

```kotlin
annotation class Ann
class Example(
    @field : Ann
    val foo : String,
    @get : Ann
    val bar : String
) {
}
```

traceyyoshima:
```java
    @Test
    void getUseSiteOnConstructorParams() {
        rewriteRun(
          kotlin("@Repeatable annotation class A"),
          kotlin(
            """
              class Example( @get : A val foo : String , @get : A val bar : String )
              """
          )
        );
    }
```

This section of code needs to be changed: https://github.com/openrewrite/rewrite-kotlin/blob/main/src/main/java/org/openrewrite/kotlin/internal/KotlinParserVisitor.java#L3420.

Pseudo code:
```
// Change this to iterate through value parameters 
convertAll(firPrimaryConstructor.getValueParameters(), commaDelim, t -> sourceBefore(")"), ctx, true), Markers.EMPTY)

// In the loop, detect annotations via PSI at the position of the parameter:
int beforeCursor = cursor;
whitespace();
PsiElement sourceElement = getCurrentPsiNode()
cursor = beforeCursor;

// if ANNOTATION_ENTRY is detected ...
// find FirProperty that is injected by the compiler in membersMultiVariablesSeparated.
// The property will have a `source.kind` of `KtFakeSourceElementKind$PropertyFromParameter`.
// The fake source and the primary constructor parameter will share the same start offset.
// The fake source will have an annotation at the location of the use site.
```

PR #175, opened 2023-07-12
Fix parsing error of annotations on constructor parameters

kunli2:
Fixes https://github.com/openrewrite/rewrite-kotlin/issues/173


kunli2:
closing this one since we don't want to parse `fake` kind FIR elements to prevent something broken.

PR #177, opened 2023-07-13
Fix paring error of annotations on constructor parameters - 2

### LINKED ISSUE #173: Parsing error of constructor parameters with annotation

kunli2:
Fixes #173 .

https://github.com/openrewrite/rewrite-kotlin/pull/175 is the 1st approach to fix this problem. 
Discussed with @traceyyoshima that one might break something since we should not parse `fake` kind FIR element (generated by the compiler but not in the source code).
This is a 2nd approach to fix #173

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/internal/KotlinParserVisitor.java
+++ b/src/main/java/org/openrewrite/kotlin/internal/KotlinParserVisitor.java
@@ -2737,14 +2737,8 @@ public class KotlinParserVisitor extends FirDefaultVisitor<J, ExecutionContext> 
         List<FirAnnotation> firAnnotations = valueParameter.getAnnotations();
         if (generatedFirProperties.containsKey(range)) {
             FirProperty generatedFirProperty = generatedFirProperties.get(range);
-            if (generatedFirProperty.getGetter() != null &&
-                !generatedFirProperty.getGetter().getAnnotations().isEmpty()) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/kotlin/internal/KotlinParserVisitor.java
```diff
@@ -3451,6 +3467,9 @@ public J visitRegularClass(FirRegularClass regularClass, ExecutionContext ctx) {
                 jcEnums.add(declaration);
             } else if (declaration instanceof FirPrimaryConstructor) {
                 firPrimaryConstructor = (FirPrimaryConstructor) declaration;
+            } else if (declaration instanceof FirProperty && declaration.getSource().getKind() instanceof KtFakeSourceElementKind.PropertyFromParameter) {
```
traceyyoshima, line 3470: getSource on FirElements is nullable, so we need to check `declaration.getSource() != null`

traceyyoshima (approved):
Interesting solution, nice job.
Minor comment to prevent NPE.

src/main/java/org/openrewrite/kotlin/internal/KotlinParserVisitor.java
kunli2: revised.

Issue #176, opened 2023-07-12
Revise visitProperty to use Source.Kind for detecting real getter and setters.


traceyyoshima:
A property getter and setter will exist in the FIR either as a Fake source generated by a use-site annotation or as a declaration.

The current detection for getters and setters will break if annotations are declared on the annotation.
Now that we know how to detect the `real sources` vs. `fake sources` and how fake sources are generated, we can safely revise the detection method.

Revise https://github.com/openrewrite/rewrite-kotlin/blob/main/src/main/java/org/openrewrite/kotlin/internal/KotlinParserVisitor.java#L1815 to check for the existence of getters and setters that are not KtRealSourceElementKind.

PR #178, opened 2023-07-13
Fix parsing error of annotation on explicit getter/setter

kunli2:
Fix parsing error of annotation on explicit getter/setter

also made an enhancement to
fixes https://github.com/openrewrite/rewrite-kotlin/issues/176

A failing test:

```java
    @Test
    void annotationOnExplicitGetter() {
        rewriteRun(
          kotlin(
            """
              annotation class A
              class Test {
                  public var stringRepresentation : String = ""
                      @A
                      get ( ) = field
                      @A
                      set ( value ) {
                          field = value
                      }
              }
              """
          )
        );
    }
```

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/internal/KotlinParserVisitor.java
+++ b/src/main/java/org/openrewrite/kotlin/internal/KotlinParserVisitor.java
@@ -1809,26 +1809,72 @@ public class KotlinParserVisitor extends FirDefaultVisitor<J, ExecutionContext> 
                 }
 
                 int saveCursor = cursor;
-                whitespace();
+                Space s0 = whitespace();
 
+                PsiElement currentPsiElement = getCurrentPsiNode();
+                KtModifierList modifierList = null;
+                List<J.Annotation> annos = new ArrayList<>();
+
+                if (currentPsiElement instanceof KtAnnotationEntry) {
+                    modifierList = PsiTreeUtil.getParentOfType(currentPsiElement, KtModifierList.class);
+                }
+
+                if (modifierList != null) {
+                    List<FirAnnotation> firAnnotations = new ArrayList<>();
+                    firAnnotations.addAll(property.getAnnotations());
+                    firAnnotations.addAll(property.getGetter().getAnnotations());
+                    firAnnotations.addAll(property.getSetter().getAnnotations());
+                    mapModifierList(modifierList, firAnnotations, annos);
+                    if (!annos.isEmpty()) {
+                        J.Annotation a0 = annos.get(0).withPrefix(s0);
+                        annos.set(0, a0);
+                    }
+                }
+
+                boolean hasAnno = !annos.isEmpty();
+                if (hasAnno) {
+                    saveCursor = cursor;
+                } else {
+                    cursor(saveCursor);
+                }
+
+                Space s1 = whitespace();
                 String methodName = source.substring(cursor, cursor + 3);
                 switch (methodName) {
                     case "get":
-                        cursor(saveCursor);
+                        if (!hasAnno) {
+                            cursor(saveCursor);
+                        }
                         if (isValidGetter(property.getGetter())) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/kotlin/internal/KotlinParserVisitor.java
```diff
@@ -1809,56 +1809,145 @@ public J visitProperty(FirProperty property, ExecutionContext ctx) {
...
+                PsiElement currentPsiElement = getCurrentPsiNode();
+                KtModifierList modifierList = null;
+                List<J.Annotation> annos = new ArrayList<>();
+
+                if (currentPsiElement instanceof KtAnnotationEntry) {
+                    modifierList = PsiTreeUtil.getParentOfType(currentPsiElement, KtModifierList.class);
+                }
+
+                if (modifierList != null) {
+                    List<FirAnnotation> firAnnotations = new ArrayList<>();
```
traceyyoshima, line 1823: Minor: size can be set ahead of time.

src/main/java/org/openrewrite/kotlin/internal/KotlinParserVisitor.java
```diff
@@ -1809,56 +1809,145 @@ public J visitProperty(FirProperty property, ExecutionContext ctx) {
...
+                    firAnnotations.addAll(property.getAnnotations());
+                    firAnnotations.addAll(property.getGetter().getAnnotations());
+                    firAnnotations.addAll(property.getSetter().getAnnotations());
+                    mapModifierList(modifierList, firAnnotations, annos);
+                    if (!annos.isEmpty()) {
+                        J.Annotation a0 = annos.get(0).withPrefix(s0);
```
traceyyoshima, line 1829: Is there a reason to save s0 and a0 instead of resetting the cursor to savedCursor after the current PsiElement is found?

src/main/java/org/openrewrite/kotlin/internal/KotlinParserVisitor.java
```diff
@@ -1815,64 +1820,58 @@ public J visitProperty(FirProperty property, ExecutionContext ctx) {
...
+                        m = m.withParameters(ListUtils.concat(receiver, m.getParameters()));
+                    }
+                    if (hasAnnotationBeforeAccessors) {
+                        m = m.withLeadingAnnotations(annos)
+                            .withName(m.getName().withPrefix(accessorPrefix))
+                            .withPrefix(EMPTY);
+                    }
+                    expressions.add(m);
+
+                    cursor(ktPropertyAccessor.getTextRange().getEndOffset());
```
traceyyoshima, line 1874: I think setting the cursor based on the PSI may cause code to be skipped if we don't parse the elements perfectly.
Are you sure it's safe to set the cursor based on the text range?
Is there a more reliable way to increment the cursor?

src/main/java/org/openrewrite/kotlin/internal/KotlinParserVisitor.java
```diff
@@ -1815,64 +1820,58 @@ public J visitProperty(FirProperty property, ExecutionContext ctx) {
...
+                    .collect(toList());
+                if (accessors.size() > 2) {
+                    throw new RuntimeException("Detected oversized explicit property accessors from a property, should have setter/getter only");
+                }
 
-                int saveCursor = cursor;
-                whitespace();
+                for (KtPropertyAccessor ktPropertyAccessor : accessors) {
+                    Optional<KtDeclarationModifierList> maybeModifier = Arrays.stream(ktPropertyAccessor.getChildren())
+                        .filter(c -> c instanceof KtDeclarationModifierList)
```
traceyyoshima, line 1833: I'm unsure if these are based on your default IntelliJ settings -- so that you know, our convention is to continue with 8 spaces

traceyyoshima (commented):
Looking good so far, added a few questions/comments

traceyyoshima (approved):
Nice work - thank you

Issue #182, opened 2023-07-17
Revise PSIElement data structure to account for CRLF.


kunli2:
The FIR source start offset is based on the transformed PSI. The transformation includes removing unnecessary parens and excluding the CR from CRLF (caused by lexer).

The bug will apply to every location in the code that uses the FIR source startOffset, and each will need to be addressed.

A test example
```java
    @Test
    void singleVariableDeclarationWithCRLF() {
        rewriteRun(
          kotlin(
            "\r\n" +
            "val a = 1")
        );
    }
```

kunli2:
the root cause of this issue:

For the source code 

```kotlin
\r\n
val a = 1
```

we can see the literal `1` is at position 10, however, when putting a breakpoint at visitConstExpression, we will see startOffset of `1` is 9, but not 10, which means the index here is shifted (I think the compiler skipped the ERROR_ELEMENT range),
 ```
------------
Source code with index:

0
val a = 1
        1
`
```

```
------------
PSI Tree All
\---(0,11) | kotlin.FILE | KtFile | Text: "\r\nval a = 1"
    |---(0,0) | PACKAGE_DIRECTIVE | KtPackageDirective | Text: ""
    |---(0,0) | IMPORT_LIST | KtImportList | Text: ""
    |---(0,1) | ERROR_ELEMENT | PsiErrorElementImpl | Text: "\r"
    |   \---(0,1) | BAD_CHARACTER | LeafPsiElement | Text: "\r"
    |---(1,2) | WHITE_SPACE | PsiWhiteSpaceImpl | Text: "\n"
    \---(2,11) | PROPERTY | KtProperty | Text: "val a = 1"
        |---(2,5) | val | LeafPsiElement | Text: "val"
        |---(5,6) | WHITE_SPACE | PsiWhiteSpaceImpl | Text: " "
        |---(6,7) | IDENTIFIER | LeafPsiElement | Text: "a"
        |---(7,8) | WHITE_SPACE | PsiWhiteSpaceImpl | Text: " "
        |---(8,9) | EQ | LeafPsiElement | Text: "="
        |---(9,10) | WHITE_SPACE | PsiWhiteSpaceImpl | Text: " "
        \---(10,11) | INTEGER_CONSTANT | KtConstantExpression | Text: "1"
            \---(10,11) | INTEGER_LITERAL | LeafPsiElement | Text: "1"
```

[...]

PR #183, opened 2023-07-17
Fix parsing error of CRLF before a variable declaration

kunli2:
Fixes https://github.com/openrewrite/rewrite-kotlin/issues/182
The root cause is [here](https://github.com/openrewrite/rewrite-kotlin/issues/182#issuecomment-1638870663)
The solution is never to use ASTNode offset on source code since it might be shifted.


traceyyoshima (approved):
Nice work Kun, you're understanding how the parser works :)

PR #185, opened 2023-07-20
Populate `fieldType`

kunli2:
`filedType` is missing, so add it in the same way as how the Java parser handles it.


src/main/java/org/openrewrite/kotlin/KotlinTypeGoat.kt
```diff
@@ -19,12 +19,16 @@ package org.openrewrite.kotlin
 
 import java.lang.Object
 
+// Whenever this class is changed, make a corresponding change in KotlinTypeGoat in the main resources folder.
```
traceyyoshima, line 22: FYI I added a comment that was missing from JavaTypeGoat to ensure the two files are updated together.

src/test/java/org/openrewrite/kotlin/.editorconfig
```diff
@@ -2,4 +2,5 @@ root = true
 
 [*.java]
 indent_size = 4
-ij_continuation_indent_size = 2
+ij_continuation_indent_size = 4
+# ij_continuation_indent_size = 8
```
traceyyoshima, line 6: I added a line when applying formatting to TypeMapping and SignatureBuilding tests.

src/test/java/org/openrewrite/kotlin/KotlinTypeMappingTest.java
```diff
@@ -62,6 +77,17 @@ void extendsKotlinAny() {
         assertThat(goatType.getSupertype().getFullyQualifiedName()).isEqualTo("kotlin.Any");
     }
 
+    @Test
+    void fieldType() {
+        J.VariableDeclarations.NamedVariable variable = getField("field").getVariables().get(0);
+        J.Identifier id = variable.getName();
+        assertThat(variable.getType()).isEqualTo(id.getType());
```
traceyyoshima, line 84: Info on test update:
The variable and name should have the same type
The fieldType should be a JavaType.Variable.
Updated to check the field type and type signatures.

src/test/java/org/openrewrite/kotlin/KotlinTypeSignatureBuilderTest.java
```diff
@@ -147,6 +147,12 @@ void parameterizedField() {
                 .isEqualTo("org.openrewrite.kotlin.KotlinTypeGoat{name=parameterizedField,type=org.openrewrite.kotlin.PT<org.openrewrite.kotlin.KotlinTypeGoat$TypeA>}");
     }
 
+    @Test
+    void fieldType() {
```
traceyyoshima, line 151: Added signature test.

traceyyoshima (approved):
Nice work Kun!

I checked the types and compared them to Java, the field type types is set correctly.

src/test/java/org/openrewrite/kotlin/.editorconfig
```diff
@@ -3,3 +3,4 @@ root = true
 [*.java]
 indent_size = 4
 ij_continuation_indent_size = 2
+# ij_continuation_indent_size = 8
```
traceyyoshima, line 6: I added an option that may be uncommented to apply formatting to TypeMapping and SignatureBuilder tests.

The config only exists to accelerate implementing `RewriteTest` formatting. Standard Java formatting still applies to other Java code.

Issue #201, opened 2023-07-25
Parsing error of binary type with `!in` operator


kunli2:
Here is an example
```java
    @Test
    void notInCondition() {
        rewriteRun(
          kotlin(
            """
              fun test() {
                  val numbers = listOf( 1, 2, 3)
                  val target = 1
                  if (target !in numbers) {
                  }
              }
              """
          )
        );
    }
```

PR #202, opened 2023-07-25
Fix `!in` binary operator parsing

kunli2:
fixes https://github.com/openrewrite/rewrite-kotlin/issues/201

I am just aware that `!in` is a binary operator in Kotlin, and it's debug name is `not`.
however, `not` is parsed to a unary currently and caused an error.

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/internal/KotlinParserVisitor.java
+++ b/src/main/java/org/openrewrite/kotlin/internal/KotlinParserVisitor.java
@@ -20,6 +20,7 @@ import org.jetbrains.kotlin.com.intellij.lang.ASTNode;
 import org.jetbrains.kotlin.com.intellij.openapi.util.TextRange;
 import org.jetbrains.kotlin.com.intellij.psi.PsiElement;
 import org.jetbrains.kotlin.com.intellij.psi.impl.source.tree.LeafPsiElement;
+import org.jetbrains.kotlin.com.intellij.psi.tree.IElementType;
 import org.jetbrains.kotlin.com.intellij.psi.util.PsiTreeUtil;
 import org.jetbrains.kotlin.descriptors.ClassKind;
 import org.jetbrains.kotlin.descriptors.annotations.AnnotationUseSiteTarget;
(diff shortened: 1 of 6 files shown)


traceyyoshima (changes requested):
Hi Kun, it was a nice attempt. We cannot merge this — a unary not operator is not a binary expression.

I’ll show you how to assess what’s happening tomorrow.

kunli2:
Yes, a unary not operator will not be a binary expression, I think it's handled correctly, since we know it's not a binary type in the code from psi

src/main/java/org/openrewrite/kotlin/internal/KotlinParserVisitor.java
```diff
@@ -931,19 +931,36 @@ public J visitSuperReference(FirSuperReference superReference, ExecutionContext
...
+
+        KtSourceElement source = functionCall.getSource();
+        if (source == null) {
+            throw new UnsupportedOperationException("The source of a FirFunctionCall is null");
+        }
 
         J j;
-        if (origin == FirFunctionCallOrigin.Operator && !(functionCall instanceof FirImplicitInvokeCall)) {
+        String type = Objects.requireNonNull(source.getElementType()).toString();
+        if ("BINARY_EXPRESSION".equals(type) ||
```
traceyyoshima, line 945: I'd like to discuss this with @sambsnyd tomorrow before we move forward with this PR. So that we can discuss the pros and cons of the current approach and how to represent the syntax sugar of operator overloading.

My main concern is that we'll already be doing a major version bump due to significant changes and it may be best to do them all at once.

We may do something similar here to handle operator overloads by wrapping J trees in a K tree that allow data flow to work and still parse and print the source correctly.

That would involve detecting the types like BINARY_EXPRESSION and PREFIX_EXPRESSION and wrapping a K tree in the method invocations. However, I am not certain how complicated it will be. So, I'd like to discuss the pros and cons first.

src/main/java/org/openrewrite/kotlin/internal/KotlinPrinter.java
```diff
@@ -84,6 +84,9 @@ public J visitBinary(K.Binary binary, PrintOutputCapture<P> p) {
             case Contains:
                 keyword = "in";
                 break;
+            case Not:
```
traceyyoshima, line 87: We'll need to adjust binary types based on what we decide during the discussion.
I think it will make more sense to set the type based on the keywords rather than the desugared syntax.

I.E. `Contains` => `In` and `Not` => `NotIn`.

[Source context](https://kotlinlang.org/docs/operator-overloading.html#in-operator)
`a in b` is translated to `b.contains(a)` and `a !in b` is translated to `!b.contains(a)`.

So it would either be the binary name `NotIn` or the translated name `NotContains`.

src/test/java/org/openrewrite/kotlin/tree/WhenTest.java
```diff
@@ -84,7 +84,7 @@ void inRange() {
           kotlin(
             """
               fun method ( i : Int ) : String {
-                  when ( i ) {
+                  return when ( i ) {
```
traceyyoshima, line 87: Nice catch - would you mind adding this fix as a separate push to main?

src/main/java/org/openrewrite/kotlin/internal/KotlinParserVisitor.java
```diff
@@ -1434,6 +1451,23 @@ private J mapKotlinBinaryOperation(FirFunctionCall functionCall) {
                 right = convertToExpression(rhs, ctx);
 
                 break;
+            }
+            case "not": {
+                LinkedHashMap<FirExpression, FirValueParameter> mapping = ((FirResolvedArgumentList) (((FirFunctionCall) (functionCall.getExplicitReceiver())).getArgumentList())).getMapping();
+                FirExpression lhs = mapping.keySet().stream().findFirst().orElse(null);
```
traceyyoshima, line 1457: I'd prefer to use a loop here instead of a stream.

src/main/java/org/openrewrite/kotlin/internal/KotlinParserVisitor.java
```diff
@@ -1434,6 +1451,23 @@ private J mapKotlinBinaryOperation(FirFunctionCall functionCall) {
...
+                if (lhs instanceof FirWhenSubjectExpression) {
+                    left = new J.Empty(randomId(), EMPTY, Markers.EMPTY);
+                } else {
+                    left = lhs != null ? convertToExpression(lhs, ctx) : null;
+                }
+
+                opPrefix = sourceBefore("!in");
+                kotlinBinaryType = K.Binary.Type.Not;
+
+                FirExpression rhs = ((FirFunctionCall)functionCall.getExplicitReceiver()).getExplicitReceiver();
```
traceyyoshima, line 1467: The explicit receiver is nullable, is this safe?

traceyyoshima (changes requested):
Nice progress Kun, I'd like to wait on this PR until we discuss how to approach operator overloads.

PR #206, opened 2023-07-31
Add spaces visitor

kunli2:
Introduce a space visitor to support all Kotlin specific space code styles.


src/main/java/org/openrewrite/kotlin/format/Spaces.java
```diff
@@ -0,0 +1,66 @@
...
+import org.openrewrite.internal.lang.Nullable;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaSourceFile;
+import org.openrewrite.kotlin.KotlinIsoVisitor;
+import org.openrewrite.kotlin.style.IntelliJ;
+import org.openrewrite.kotlin.style.SpacesStyle;
+
+import static java.util.Objects.requireNonNull;
+
+public class Spaces extends Recipe {
```
sambsnyd, line 28: We don't need a recipe just for spaces, we need a full `AutoFormat`. This should be the start of the kotlin `AutoFormat`.

src/main/java/org/openrewrite/kotlin/format/Spaces.java
```diff
@@ -0,0 +1,66 @@
...
+        public @Nullable J visit(@Nullable Tree tree, ExecutionContext executionContext) {
+            if (tree instanceof JavaSourceFile) {
+                JavaSourceFile cu = (JavaSourceFile) requireNonNull(tree);
+                SpacesStyle style = ((SourceFile) cu).getStyle(SpacesStyle.class);
+                if (style == null) {
+                    style = IntelliJ.spaces();
+                }
+                doAfterVisit(new SpacesVisitor<>(style));
+            }
+            return super.visit(tree, executionContext);
```
sambsnyd, line 56: Wasteful to continue to visit the sub-elements of the current tree

src/main/java/org/openrewrite/kotlin/format/Spaces.java
```diff
@@ -0,0 +1,66 @@
...
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new SpacesFromCompilationUnitStyle();
+    }
+
+    private static class SpacesFromCompilationUnitStyle extends KotlinIsoVisitor<ExecutionContext> {
+        @Override
+        public @Nullable J visit(@Nullable Tree tree, ExecutionContext executionContext) {
+            if (tree instanceof JavaSourceFile) {
+                JavaSourceFile cu = (JavaSourceFile) requireNonNull(tree);
```
sambsnyd, line 49: `instanceof` check guarantees non-null already, redundant to also call `requireNonNull`

src/main/java/org/openrewrite/kotlin/format/Spaces.java
sambsnyd, line 56:
```suggestion
            if (!(tree instanceof K.CompilationUnit)) {
                return tree;
            }
            K.CompilationUnit cu = (K.CompilationUnit) tree;
            SpacesStyle style = cu.getStyle(SpacesStyle.class, IntelliJ.spaces());
            return new SpacesVisitor<>(style).visitNonNull(cu, getCursor().fork());
```

src/main/java/org/openrewrite/kotlin/format/Spaces.java
```diff
@@ -0,0 +1,66 @@
...
+            return super.visit(tree, executionContext);
+        }
+    }
+
+    public static <J2 extends J> J2 formatSpaces(J j, Cursor cursor) {
+        SourceFile cu = cursor.firstEnclosingOrThrow(SourceFile.class);
+        SpacesStyle style = cu.getStyle(SpacesStyle.class);
+        //noinspection unchecked
+        return (J2) new SpacesVisitor<>(style == null ? IntelliJ.spaces() : style).visitNonNull(j, 0, cursor);
+    }
```
sambsnyd, line 65: We don't need this method.
Instead `KotlinVisitor` should override `autoformat()` and `maybeAutoformat()` methods from `JavaVisitor` to invoke kotlin auto format

src/main/java/org/openrewrite/kotlin/format/SpacesVisitor.java
```diff
@@ -0,0 +1,1232 @@
...
+
+    @Override
+    public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration classDecl, P p) {
+        J.ClassDeclaration c = super.visitClassDeclaration(classDecl, p);
+        if (c.getBody() != null) {
+            c = c.withBody(spaceBefore(c.getBody(), beforeLeftBrace));
+            if (c.getBody().getStatements().isEmpty()) {
+                if (c.getKind() != J.ClassDeclaration.Kind.Type.Enum) {
+                    // Defaulted to `false` in IntelliJ's Kotlin formatting.
+                    boolean withinCodeBraces = false;
```
sambsnyd, line 211: These should be hardcoded, not read from the style

src/main/java/org/openrewrite/kotlin/internal/KotlinPrinter.java
```diff
@@ -858,10 +858,6 @@ public J visitTernary(J.Ternary ternary, PrintOutputCapture<P> p) {
         @Override
         public J visitTypeCast(J.TypeCast typeCast, PrintOutputCapture<P> p) {
             beforeSyntax(typeCast, Space.Location.TYPE_CAST_PREFIX, p);
-            boolean printParens = !typeCast.getMarkers().findFirst(OmitParentheses.class).isPresent();
-            if (printParens) {
-                p.append('(');
-            }
```
sambsnyd, line 864: How does this not break kotlin code that uses lambdas?
```kotlin
listOf(1,2,3).forEach { println(it) } // very common to omit parentheses here
```

src/main/java/org/openrewrite/kotlin/format/Spaces.java
kunli2: I see, it should be
```java 
return (J) tree; 
```

src/main/java/org/openrewrite/kotlin/format/Spaces.java
kunli2: I think this file is copied from `rewrite`, I checked the code in the rewrite, and it seems this is only used for the convenience of test purposes. I feel we should remove `Spaces` recipe but keep the internal static class `SpacesFromCompilationUnitStyle` which includes style detector.

src/main/java/org/openrewrite/kotlin/format/Spaces.java
kunli2: revised.

src/main/java/org/openrewrite/kotlin/format/SpacesVisitor.java
kunli2: Sure, I am going to clean up these kind of hard-coded stuffs after moving all space tests to `rewrite-kotlin`

src/main/java/org/openrewrite/kotlin/format/Spaces.java
kunli2: yes, this is copied from rewrite and not used anywhere, removed.

src/main/java/org/openrewrite/kotlin/internal/KotlinPrinter.java
kunli2: I am not sure I understand this question, 
I think for lambda, it's handled by `visitLambda` method in this file.

src/main/java/org/openrewrite/kotlin/format/Spaces.java
kunli2: removed this file, and have `SpacesFromCompilationUnitStyle` instead.

src/main/java/org/openrewrite/kotlin/format/Spaces.java
kunli2: We haven't had an `AutoFormatVisitor` yet but just introduced SpacesVisitor only in this PR. 
will make the change of override `autoformat()` and `maybeAutoformat()` methods in `KotlinVisitor` in a separate PR

src/main/java/org/openrewrite/kotlin/internal/KotlinPrinter.java
sambsnyd: Oh yeah this is type casts.
In kotlin there are two kinds of type casts, the java-style `(Foo) bar` and `bar as Foo`. Does the latter still print correctly with this change?

kunli2: Yeah, this code is actually removed from @traceyyoshima  initial change, he might have more context about why remove this, but I verified the printer works for the latter for the below case. 
```kotlin
fun main() {
    val obj: Any = "Hello"
    val str: String = obj as String
}
```
but I am not sure how to do java-style type case in Kotlin.
I tried this but doesn't work
```kotlin
fun main() {
    val obj: Any = "Hello"
    // val str: String = obj as String
    val str2: String = (String) obj
}
```

PR #213, opened 2023-08-09
New `AddImport` visitor

knutwannheden:
The Java `AddImport` visitor doesn't quite fit the bill for Kotlin, because Kotlin doesn't really have static imports (although companion members as well as static Java members can be imported using the regular import syntax) and it has the additional concept of aliases for imports (also called renames in some languages). Additionally, the imports are typically grouped differently in Kotlin.

This PR provides a Kotlin-specific `AddImport` visitor which can be used in the same way as the Java-specific visitor (via `maybeAddImport()` on the base visitor class.


knutwannheden:
@sambsnyd This is still work in progress and I've created the PR as a draft for now.

sambsnyd (approved):

PR #221, opened 2023-08-15
Add RemoveTrailingSemicolon recipe

kunli2:
Add a recipe `RemoveTrailingSemicolon` to remove redundant trailing semicolons in Kotlin, and it looks like this is common mistake from Java developers.

Also, update `KotlinPrinter` to be extendable.

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/cleanup/RemoveTrailingSemicolon.java
+++ b/src/main/java/org/openrewrite/kotlin/cleanup/RemoveTrailingSemicolon.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 the original author or authors.
+ * Copyright 2023 the original author or authors.
  * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


src/main/java/org/openrewrite/kotlin/cleanup/RemoveTrailingSemicolon.java
```diff
@@ -0,0 +1,176 @@
+/*
+ * Copyright 2021 the original author or authors.
```
traceyyoshima, line 2:
```suggestion
 * Copyright 2023 the original author or authors.
```

src/test/java/org/openrewrite/kotlin/cleanup/RemoveTrailingSemicolonTest.java
```diff
@@ -0,0 +1,193 @@
+/*
+ * Copyright 2021 the original author or authors.
```
traceyyoshima, line 2:
```suggestion
 * Copyright 2023 the original author or authors.
```

knutwannheden:
There are still esoteric cases when semicolons don't get removed. E.g. when there is a comment at the end of the line after the semicolon. But I think the recipe is OK as it is now.

knutwannheden (approved):

Issue #220, opened 2023-08-14
Parsing error with trailing comma in an array declaration


kunli2:
Parsing error with trailing comma in an array declaration.
 
An example
```
    @Test
    void trailingComma() {
        rewriteRun(
          kotlin(
            """
              val integerArray: Array<Int> = arrayOf(
                  1,
                  2,
                  3,
              )
              """
          )
        );
    }
```

Issue #216, opened 2023-08-11
Enum class parsing error if it has a trailing comma


kunli2:
Enum class parsing error if it has a trailing comma

An example:
```java
    @Test
    void comma() {
        rewriteRun(
          kotlin(
            """
              enum class Scope {
                  None,
                  Compile,
              }
              """
          )
        );
    }
```

PR #222, opened 2023-08-15
Kotlin allows trailing commas in method invocations and after enums

knutwannheden:
The OpenRewrite Kotlin parser saves these in `Space` elements which in Java typically only contain whitespace characters.

Fixes: #220, #216

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/Assertions.java
+++ b/src/main/java/org/openrewrite/kotlin/Assertions.java
@@ -20,13 +20,13 @@ import org.intellij.lang.annotations.Language;
 import org.openrewrite.ExecutionContext;
 import org.openrewrite.SourceFile;
 import org.openrewrite.internal.lang.Nullable;
-import org.openrewrite.java.tree.J;
-import org.openrewrite.java.tree.Space;
+import org.openrewrite.java.tree.*;
 import org.openrewrite.kotlin.internal.KotlinParsingException;
 import org.openrewrite.kotlin.tree.K;
 import org.openrewrite.test.SourceSpec;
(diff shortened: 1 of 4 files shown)


knutwannheden:
Note: AFAICT this is handled the same way in Java (i.e. using `Space#whitespace`).

traceyyoshima:
> Note: AFAICT this is handled the same way in Java (i.e. using `Space#whitespace`).

I think this is caused by an assumption in container conversion where there is likely a `sourceBefore(")")`.
Rather than adding the trailing comma to whitespace, it may be best to use the `TrailingComma` markers and store the suffix separately.

knutwannheden:
> I think this is caused by an assumption in container conversion where there is likely a `sourceBefore(")")`.
> Rather than adding the trailing comma to whitespace, it may be best to use the `TrailingComma` markers and store the suffix separately.

Yes, that sounds great. I was actually looking for this marker, but apparently in the wrong place (thought it would be in `rewrite-kotlin`, but of course `rewrite-java` makes more sense).

knutwannheden:
Makes me wonder if we shouldn't also apply this marker in Java (array initializers and enums also allow this).

knutwannheden:
Fixed with https://github.com/openrewrite/rewrite-kotlin/pull/229 and https://github.com/openrewrite/rewrite-kotlin/commit/2ee58f749b28ff409da7a5b588fa0cbfe1382211

traceyyoshima:
> Makes me wonder if we shouldn't also apply this marker in Java (array initializers and enums also allow this).

Agreed, the TrailingComma marker is something each parser could use and respond to based on conventions.

PR #223, opened 2023-08-15
Fix printing of `return` with label

knutwannheden:


src/main/java/org/openrewrite/kotlin/internal/KotlinPrinter.java
```diff
@@ -162,11 +162,17 @@ public J visitFunctionType(K.FunctionType functionType, PrintOutputCapture<P> p)
     @Override
     public J visitKReturn(K.KReturn kReturn, PrintOutputCapture<P> p) {
         visit(kReturn.getAnnotations(), p);
-        visit(kReturn.getExpression(), p);
+        J.Return return_ = kReturn.getExpression();
+        beforeSyntax(return_, Space.Location.RETURN_PREFIX, p);
+        p.append("return");
```
traceyyoshima, line 167: A quick note here:
KReturn exists to support labels, and J.Return is still a part of the tree since K extends J.

In Kotlin, the `return` keyword is not required since it may be implied. So, we use a marker to prevent the print when applicable: https://github.com/openrewrite/rewrite-kotlin/blob/8499d1f769d3664dcc38b4942f88b47839603d75/src/main/java/org/openrewrite/kotlin/internal/KotlinPrinter.java#L834.

The return keyword is required when a label exists, so we may want to change KReturn to have a non-nullable Label so that implicit returns only happen on J.Return.

knutwannheden: @traceyyoshima Thanks for clarifying. I think that would probably simplify things. We will however still have to inline the printing of the `return` keyword, unless we in the `KotlinJavaPrinter` want to use the cursor (or cursor messaging) to retrieve the label.

I can look into implementing what you suggest.

knutwannheden: @traceyyoshima If we now change the `K.KReturn#label` to be non-nullable, then we will still have to expect `null` values for a while, as old LST files will still have them. So I am a bit uncertain if we really want to do that.

traceyyoshima (approved):

PR #225, opened 2023-08-16
Correct printing of `!in` expressions

knutwannheden:
A Kotlin binary expression using the `!in` operator is parsed as a `K.Binary` (with operator `Contains`) inside a `J.Unary` (with operator `Not`). The `KotlinPrinter` prints a leading `!` back out, so `a !in b` is printed as `!a !in b`.

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/format/MinimumViableSpacingVisitor.java
+++ b/src/main/java/org/openrewrite/kotlin/format/MinimumViableSpacingVisitor.java
@@ -203,7 +203,7 @@ public class MinimumViableSpacingVisitor<P> extends KotlinIsoVisitor<P> {
     @Override
     public K.Binary visitBinary(K.Binary binary, P p) {
         K.Binary kb = super.visitBinary(binary, p);
-        if (kb.getOperator() ==  K.Binary.Type.Contains) {
+        if (kb.getOperator() ==  K.Binary.Type.Contains || kb.getOperator() ==  K.Binary.Type.NotContains) {
             kb = kb.getPadding().withOperator(kb.getPadding().getOperator().withBefore(updateSpace(kb.getPadding().getOperator().getBefore(), true)));
             kb = kb.withRight(spaceBefore(kb.getRight(), true));
         }
--- a/src/main/java/org/openrewrite/kotlin/format/SpacesVisitor.java
+++ b/src/main/java/org/openrewrite/kotlin/format/SpacesVisitor.java
@@ -764,6 +764,7 @@ public class SpacesVisitor<P> extends KotlinIsoVisitor<P> {
         K.Binary b = super.visitBinary(binary, p);
         K.Binary.Type operator = b.getOperator();
         switch (operator) {
+            case NotContains:
             case Contains:
                 break;
             case Get:
--- a/src/main/java/org/openrewrite/kotlin/internal/KotlinPrinter.java
+++ b/src/main/java/org/openrewrite/kotlin/internal/KotlinPrinter.java
@@ -91,6 +91,9 @@ public class KotlinPrinter<P> extends KotlinVisitor<PrintOutputCapture<P>> {
             case Contains:
                 keyword = "in";
                 break;
+            case NotContains:
+                keyword = "!in";
+                break;
             case IdentityEquals:
                 keyword = "===";
                 break;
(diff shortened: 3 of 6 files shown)


knutwannheden:
@traceyyoshima Not sure if we should do it like I do it in this PR or if we should introduce a `NotContains` operator or yet some other way (marker?).

Also note that the printer currently relies on the `Space` to print out the `!` of `!in`. This doesn't feel quite right as constructing this LST "manually" would require adding this `Space` correctly. Should we instead have a marker for this or "derive" it from the owning `J.Unary`?

src/main/java/org/openrewrite/kotlin/Assertions.java
```diff
@@ -99,7 +99,8 @@ public static Consumer<SourceSpec<K.CompilationUnit>> isFullyParsed() {
             new KotlinIsoVisitor<Integer>() {
                 @Override
                 public Space visitSpace(Space space, Space.Location loc, Integer integer) {
-                    if (!space.getWhitespace().trim().isEmpty()) {
+                    String trimmed = space.getWhitespace().trim();
+                    if (!trimmed.isEmpty() && !trimmed.equals("!")) {
```
knutwannheden, line 103: Now that we have this generic check to verify if the printer prints back out whatever was parsed, I am a bit unsure if we still need this `isFullyParsed()` assertion or not. If we still want to keep it, I would change this line to only allow the `!` when we are in a context where this is actually allowed.

traceyyoshima: I think it's fine to remove `isFullyParsed()`. It was easier to find parsing issues before the idempotent print check was added because we add parser error information when it happens. Most of that is lost now, so we can remove the J.Unknown as well.

traceyyoshima:
For context, in case you don't know yet:
The code `n !in 1 .. 4` is sugar for operator overloads ([info on overloads](https://kotlinlang.org/docs/operator-overloading.html#binary-operations)). We discussed Kotlin sugar and the de-sugaring done by the compiler with Jon and plan to use `J` trees as much as possible to enable data flow support.

This is an example of what is actually happening:
```kotlin
fun method ( n : Int ) {
    // syntax sugar for `rangeTo`, `contains`, and `not`
    if ( n !in 1 .. 4 ) {}

    // syntax sugar for `not`
    if ( ! 1 . rangeTo ( 4 ) . contains( n ) ) {}

    // the full method invocation chain that we visit in the Kotlin compiler output.
    if ( 1 . rangeTo ( 4 ) . contains ( n ) . not ( ) ) {}
}
```

A full method invocation chain like `if ( 1 . rangeTo ( 4 ) . contains ( n ) . not ( ) ) {}` will be naturally visited in a different way. So, we really only need to handle the sugar/desugared code.

In the case of `n !in 1 .. 4`, we would use the compiler trees to create the method invocation changes and preserve them in a KBinary. I finished most of the work, and ran out of time before checking for edge cases and cleaning up the code. I pushed the changes [here](https://github.com/openrewrite/rewrite-kotlin/commit/915eb8719d1a2d3bfce4657a303a45acce2be71d).

The operator sugar is `!in`, so I think it's appropriate to add a `NotContains`.
For the sake of velocity, you can add `NotContains` and I'll pick up on the changes when I have time.
Unless you'd like to take over the branch.

traceyyoshima (changes requested):
@knutwannheden For context, in case you don't know yet:
The code n !in 1 .. 4 is sugar for operator overloads ([info on overloads](https://kotlinlang.org/docs/operator-overloading.html#binary-operations)). We discussed Kotlin sugar and the de-sugaring done by the compiler with Jon and plan to use J trees as much as possible to enable data flow support.

This is an example of what is actually happening:

fun method ( n : Int ) {
    // syntax sugar for `rangeTo`, `contains`, and `not`
    if ( n !in 1 .. 4 ) {}

    // syntax sugar for `not`
    if ( ! 1 . rangeTo ( 4 ) . contains( n ) ) {}

    // the full method invocation chain that we visit in the Kotlin compiler output.
    if ( 1 . rangeTo ( 4 ) . contains ( n ) . not ( ) ) {}
}
A full method invocation chain like `if ( 1 . rangeTo ( 4 ) . contains ( n ) . not ( ) ) {}` will be naturally visited in a different way. So, we really only need to handle the sugar/desugared code.

In the case of `n !in 1 .. 4`, we would use the compiler trees to create the method invocation changes and preserve them in a KBinary. I finished most of the work, and ran out of time before checking for edge cases and cleaning up the code. I pushed the changes [here](https://github.com/openrewrite/rewrite-kotlin/commit/915eb8719d1a2d3bfce4657a303a45acce2be71d).

The operator sugar is `!in`, so I think it's appropriate to add a `NotContains`.
For the sake of velocity, you can add `NotContains` and I'll pick up on the changes when I have time.

knutwannheden:
I will try to fix this one and others tomorrow.

knutwannheden:
Merged solution with `NotContains` so that we can successfully parse (and print) `!in` expressions. This relies on a hack that will soon be removed again.

Issue #100, opened 2023-04-13
Add support for anonymous lambda arguments in super class constructors.


traceyyoshima:
```java
    @ExpectedToFail
    @Issue("https://github.com/openrewrite/rewrite-kotlin/issues/100")
    @Test
    void extension() {
        rewriteRun(
          kotlin(
            """
            abstract class Test(arg: Test.() -> Unit = {}) {
                init {
                    arg()
                }
            }
            """
          ),
          kotlin(
            """
            class ExtensionTest : Test({
                val property = "property"
            })
            """
          )
        );
    }
```

Update:
The lambda argument passed into the superclass constructor is located in the declarations of the `FirRegularClass` for `ExtensionTest`. The `FirPrimaryConstructor` contains a `delegateConstructor` with the anonymous function.

PR #238, opened 2023-08-18
Support calling super constructor with anonymous lambda argument

knutwannheden:
The result is that the `J.NewClass` has an argument list with a `J.Lambda` in it.

Fixes: #100


knutwannheden:
@traceyyoshima I can't tell if this is completely correct or not.

traceyyoshima:
@knutwannheden I need to confirm a few details about delegate constructors and will look at this one after it's converted to Kotlin.

knutwannheden:
@traceyyoshima Converted to Kotlin now.

traceyyoshima (approved):
@knutwannheden I think it's fine to merge and move forward -- there might be cases where the delegate constructor is null, but I am unsure if that will happen.

PR #243, opened 2023-08-22
Correctly handle empty parens before trailing lambda

knutwannheden:
A function which takes a single lambda argument can be called either with parentheses around the lambda, without parentheses, or with empty parenthesis followed by the lambda. This last calling style was not yet handled correctly by the parser.

To do this correctly we should probably add another marker to capture the whitespace which could be inbetween the opening and closing parenthesis.

Also, the logic in `mapFunctionalCallArguments()` is still quite convoluted and can be simplified even more.

### DIFF
--- a/src/main/kotlin/org/openrewrite/kotlin/internal/KotlinParserVisitor.kt
+++ b/src/main/kotlin/org/openrewrite/kotlin/internal/KotlinParserVisitor.kt
@@ -1121,7 +1121,7 @@ class KotlinParserVisitor(
                     receiver = getReceiver(functionCall.dispatchReceiver)
                 }
                 if (receiver != null) {
-                    var selectExpr =
+                    val selectExpr =
                         convertToExpression<Expression>(receiver, data)!!
                     val after = whitespace()
                     if (skip("?")) {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/openrewrite/kotlin/internal/KotlinParserVisitor.kt
```diff
@@ -1341,99 +1341,78 @@ class KotlinParserVisitor(
...
         val args: JContainer<Expression>
-        if (firExpressions.size == 1) {
-            val firExpression = firExpressions[0]
-            args = if (firExpression is FirVarargArgumentsExpression) {
-                val argumentsExpression = firExpressions[0] as FirVarargArgumentsExpression
+        var saveCursor = cursor
+        val containerPrefix = whitespace()
+        var parenOrBrace = source[cursor++]
+        val saveCursor2 = cursor
+        val before = whitespace()
```
traceyyoshima, line 1360: I expect the tests' changes were unnecessary, and I suspect the issue is that the whitespace in `before` is not preserved. Should the `before` be added to a `J.Empty`?

traceyyoshima (commented):
Just a quick question about whitespace

src/main/kotlin/org/openrewrite/kotlin/internal/KotlinParserVisitor.kt
knutwannheden: The changes to the existing test were unnecessary yes (just did them to fix a reported error). But the new test `trailingLambdaArgumentWithParentheses()` requires the changes made here.

Not sure if adding a `J.Empty` wouldn't break things, since there will also be another argument for the trailing lambda. But one thing which isn't yet covered (for which I assume we would want to add a dedicated marker) is if there is whitespace between the opening and closing empty parenthesis characters.

PR #246, opened 2023-08-22
Use compiler analysis to check for syntax errors

knutwannheden:
Use `AnalyzerWithCompilerReport` to find syntax errors and return a `ParseError` from parser when any errors are found.

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/KotlinParser.java
+++ b/src/main/java/org/openrewrite/kotlin/KotlinParser.java
@@ -22,7 +22,9 @@ import kotlin.jvm.functions.Function1;
 import lombok.AccessLevel;
 import lombok.RequiredArgsConstructor;
 import org.intellij.lang.annotations.Language;
+import org.jetbrains.kotlin.KtRealPsiSourceElement;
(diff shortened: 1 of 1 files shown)


knutwannheden:
@traceyyoshima This is still work in progress. As you can see there are still failing tests (all related to CRLF?) and probably needs some more work. I just thought I would share this.

traceyyoshima:
> @traceyyoshima This is still work in progress. As you can see there are still failing tests (all related to CRLF?) and probably needs some more work. I just thought I would share this.

I haven't looked into the error analysis, but I noticed you mentioned CRLF.
The PSI treats CR of `\r` as a bad character, though I wouldn't expect the analysis to fail.
There may be a missing compiler configuration since Kotlin has to compile code with CRLF.

traceyyoshima (commented):
Added a comment with info about CRLF.
I'll check if there are any missing compiler configurations when/if I have time.

traceyyoshima:
Closing. Added the syntax analysis separately -- it works appropriately now that we filter out CR characters.

PR #252, opened 2023-08-24
Support references to imported members

knutwannheden:
The imported members could be JVM statics or also on some Kotlin object.

Fixes: #251


src/main/kotlin/org/openrewrite/kotlin/internal/KotlinParserVisitor.kt
```diff
@@ -2355,6 +2357,8 @@ class KotlinParserVisitor(
                 }
             }
         }
+        if (name.isEmpty())
```
knutwannheden, line 2360: I don't quite like the logic above which attempts to consume parser tokens corresponding to the name segments. I think we should check if we can somehow determine from the AST if we can expect these tokens or not or otherwise just make sure we set the cursor to the corresponding end offset at the end.

kunli2 (approved):

PR #253, opened 2023-08-24
Improve formatting of spaces around lambdas

knutwannheden:


traceyyoshima (approved):

knutwannheden:
@kunli2 You have provided the alternatives #254 and #255 which I haven't had time to review in detail yet. I still think we should add an `Infix` marker, because as stated, the corresponding method is not necessarily an extension method, so just reusing `Extension` for this is wrong and could cause trouble down the road, when we do need to make that distinction.

What are your thoughts on this? Should we proceed with this PR or can you incorporate it into one of your two PRs?

knutwannheden:
Thanks. I will take a look at #255 then.

PR #254, opened 2023-08-25
Update autoFormat to support trailing lambda well

kunli2:
Talked with @knutwannheden about the approach  https://github.com/openrewrite/rewrite-kotlin/pull/253. 
we probably don't want to introduce a new `Infix` marker here.
This is an alternative solution to PR https://github.com/openrewrite/rewrite-kotlin/pull/253 without introducing a new `Infix` marker.

Currently, there is one test failing (see below), I think it should be covered by parsing improvement, moving the space after the comma in `e, -> println(e)`  from marker `TrailingComma.suffix` to the lambda arrow.
```java
    @Test
    void trailingLambdaWithParamTrailingComment() {
        rewriteRun(
          kotlin(
            """
              val x = listOf(1).forEach { e, -> println(e) }
              """
          )
        );
    }
```


knutwannheden:
@kunli2 Thanks for this alternative. I will try to see if I can combine the two PRs into one. While adding `Infix` adds some complexity, I think it is the right thing to do, because an infix function does not have to be an extension function and down the line we may have to keep them separate. So I will try to keep that part from my PR.

PR #257, opened 2023-08-27
Align `KotlinParser.Builder` API with `JavaParser.Builder`

knutwannheden:


knutwannheden:
A releases including this PR requries a new `rewrite` release. So I think we should still wait with merging it. Setting it as draft for now.

PR #259, opened 2023-08-29
Support calls with member references without receiver as argument

knutwannheden:
Support expressions like `Supplier(::emptyList)`.

### DIFF
--- a/src/main/kotlin/org/openrewrite/kotlin/internal/KotlinParserVisitor.kt
+++ b/src/main/kotlin/org/openrewrite/kotlin/internal/KotlinParserVisitor.kt
@@ -614,7 +614,7 @@ class KotlinParserVisitor(
         }
 
         var typeArgs: JContainer<Expression>? = null
-        if (callableReferenceAccess.typeArguments.isNotEmpty()) {
+        if (getReceiver(callableReferenceAccess) != null && callableReferenceAccess.typeArguments.isNotEmpty()) {
             typeArgs = mapTypeArguments(callableReferenceAccess.typeArguments, data)
         }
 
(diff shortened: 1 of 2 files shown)


knutwannheden:
@traceyyoshima This PR allows parsing (and printing) expressions like `Supplier(::emptyList)`, but the produced LST is still wrong and I think the LST is also wrong for input which was previously already parsed like `Supplier<List<String>> { emptyList() }`, because it essentially returns a method call without proper type attribution, whereas we should probably have gotten a `NewClass` instance or what do you think?

knutwannheden:
I used a Java decompiler to check what the Kotlin compiler does for

```kt
val s1: Supplier<List<String>> = Supplier(::emptyList)
val s2: Supplier<List<String>> = Supplier { emptyList() }

val s3 = Supplier<List<String>>(::emptyList)
val s4 = Supplier<List<String>> { emptyList() }
```
Both `s1` and `s3` are compiled to `Supplier s1 = () -> CollectionsKt.emptyList()` (generics lost due to type erasure). So it is a simple lambda (possibly with a cast, depending on context). 

For `s2` and `s4` the compiler appears to generate static helper methods and then method references to these. But this may well be different if the lambda were to reference any variables in the calling context (didn't test that).

In any case it didn't end up producing anonymous subclasses, as I would have thought.

@traceyyoshima Does the FIR contain information about the desugaring so that we know what to map it to?

knutwannheden:
I have now fixed the parsing of member references (they can never declare type parameters, AFAIK). The question regarding the mapping to `J.NewClass` with a body (and corresponding type attribution) still remains, so I leave the PR open for now.

traceyyoshima:
> I have now fixed the parsing of member references (they can never declare type parameters, AFAIK). The question regarding the mapping to J.NewClass with a body (and corresponding type attribution) still remains, so I leave the PR open for now.

Hi Knut, my apologies, I just noticed this PR. Yes, I think the J.NewClass makes the most sense too.

In regards to type attribution, the FIR typically does contain the type attribution for desugaring, however, it can be very difficult to find. I haven't found a systematic means to identify where type attribution exists in the FIR, or FirSession after desugaring.

Kotlin has many `Util`(s) for both the FIR and the PSI that have methods to access and answer questions about the tree. In the case of type attribution, most of the information exists in the FirSession, and utilities are used to access the relevant information.

Note:
Due to time constraints, we weren't allowed to spend any time in the backend phase of the compiler. The backend phase of the compiler has type mapping and signature building, so there might be useful visitors, classes, etc. I'm unsure if it's possible to use the FIR with any of the backend code since I think it uses the IR.

knutwannheden:
I will try to merge this PR later today.

knutwannheden:
Looks like everything relevant from this PR was merged already.

Issue #260, opened 2023-08-29
Parsing error


kunli2:
Parsing error when parsing below code

```kotlin
              import java.lang.annotation.RetentionPolicy

              fun addMember(name: String?, format: String?, vararg args: Any?): String? {
                  return ""
              }

              fun method(i : Int) {
                  addMember("value", "${'$'}T.${'$'}N", RetentionPolicy::class.java, "CLASS")
              }
```

PR #261, opened 2023-08-29
Fix a string template parsing issue

kunli2:
fixes https://github.com/openrewrite/rewrite-kotlin/issues/260

I have identified that 4 lines of code are the root cause of this issue, removing them resolves the issue and also passes all the tests.
However, those 4 lines of code seem to be added for a specific purpose, I would like a second opinion to review and confirm whether it is safe to remove them.


knutwannheden:
Nice. I debugged the test case and also couldn't see any reason not to remove these lines.

Issue #262, opened 2023-08-29
Alias filed access Parsing error


kunli2:
An example

```java

    @Test
    void aliasFieldAccess() {
        rewriteRun(
          kotlin(
            """
              import java.lang.Integer as Number

              class Test {
                  var max = Number.MAX_VALUE
              }
              """
          )
        );
    }
```

PR #263, opened 2023-08-30
Fix parsing error of alias import field access

kunli2:
fixes https://github.com/openrewrite/rewrite-kotlin/issues/262


knutwannheden:
@kunli2 I did a minor refactoring so that we check the import alias first.

Issue #270, opened 2023-08-31
Error parsing extension function call


knutwannheden:
The following extension function call 
```kt
val block: Collection<Any>.() -> Unit = {}
val r = listOf("descriptor").block()
```
currently prints back out as
```kt
val block: Collection<Any>.() -> Unit = {}
val r = blocklistOf("descriptor").block()
```
Most likely the parser adds some marker which then misguides the printer.

traceyyoshima:
I suspect this is caused by a condition failing to detect the select for the `block` function call.

knutwannheden:
I don't see how we can use the FIR to derive this. I will read up on https://kotlinlang.org/spec/overload-resolution.html#receivers and see if that helps.

knutwannheden:
It can be even more complicated when there are additional arguments, as the function call arguments effectively get split across three different locations:
```kt
val block2: Collection<Any>.(String, () -> Unit) -> Unit = { _, _ -> run {} }
val r2 = listOf("descriptor").block2("x") {}
```
I just now realize that the function type here in fact describes an extension function. I wasn't aware that also function types can describe extension functions. But this at least makes it consistent with the syntax and could possibly make it simpler to parse.

Makes me wonder how we account for extension functions (and what could even be anonymous functions as in `listOf("").(block2 as (Collection<Any>.(String, () -> Unit) -> Unit))("") {}`) in our `JavaType` type attribution...

PR #279, opened 2023-09-01
Fix parsing secondary constructor delegation

knutwannheden:
It is allowed for a class to not declare any primary constructor but instead secondary constructors. These can then delegate using `this` or `super`.

Additionally, this commit changes how constructors with delegation are mapped in the LST. Instead of "abusing" the return type expression (which for Java is always set to `null` for constructors) and there using a `J.NewClass` (i.e. `new`), the delegated constructor call will now instead be added as the first statement to the constructor body.

Issue: #270


knutwannheden:
@traceyyoshima This is still a draft, but maybe you could take a look anyway. I think the constructor delegation should really result in a constructor method declaration with a method invocation (corresponding to the `this()` or `super()` call) in its body. Using the return type expression seemed a bit like a hack, because in Java that is always `null` for constructors. Also this should better reflect what the compiler does (AFAICT the produced bytecode reflects what I have here).

I still leave it as a draft, because I think we probably shouldn't be using the `TypeReferencePrefix` marker anymore. Should we introduce a dedicated `ConstructorDelegation` marker to distinguish the case from a "normal" body?

knutwannheden:
I went ahead and added a `ConstructorDelegation` marker.

knutwannheden:
@traceyyoshima I now learned that constructor delegation can be combined with a body, in which case the delegation would simply be the first statement in the corresponding Java code (not even sure if Kotlin allows declaring a `super()` or `this()` call as part of the constructor body). So we would probably want to add the marker to that first statement as well and make sure it doesn't get printed back out.

src/main/java/org/openrewrite/kotlin/internal/KotlinPrinter.java
```diff
@@ -1061,6 +1069,9 @@ private void visitTypeAlias(J.VariableDeclarations typeAlias, PrintOutputCapture
 
         protected void visitStatement(@Nullable JRightPadded<Statement> paddedStat, JRightPadded.Location location, PrintOutputCapture<P> p) {
             if (paddedStat != null) {
+                if (paddedStat.getElement().getMarkers().findFirst(ConstructorDelegation.class).isPresent()) {
+                    return;
+                }
```
knutwannheden, line 1074: I am not quite happy with this part. I wonder if we should introduce an additional general purpose `Synthetic` (or `OmitStatement`, `Implied`, or similar) marker, which we additionally could add to the `this()` or `super()` call and then here check for that one instead. What do you think, @traceyyoshima ?

knutwannheden:
Apart from the comment I left regarding the marker, I feel that this PR is now ready.

knutwannheden:
@traceyyoshima I went ahead and merged the PR. We should still discuss what to do about the marker.

Issue #267, opened 2023-08-31
Error parsing annotated expressions


knutwannheden:
The following example is parsed incorrectly and leads to a parse-to-print failure:
```kt
val s = java.util.function.Supplier<String> {
    @Suppress("UNCHECKED_CAST")
    requireNotNull("x")
}
```
The reason is that the parser doesn't recognize annotations on expressions.

PR #281, opened 2023-09-04
Support annotations on expressions

knutwannheden:
There is a new `K.AnnotatedExpression` LST type which holds annotations of annotated expressions and is used to wrap expressions, whenever they are annotated in the source code.

This also required removing `K.KReturn#anntations` and because a return in Kotlin is actually an expression, `KReturn` now also implements `Expression`.

Similarly, a `K.KString` is not a statement, just an expression.

Fixes: #267

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/tree/K.java
+++ b/src/main/java/org/openrewrite/kotlin/tree/K.java
@@ -320,6 +320,66 @@ public interface K extends J {
         }
     }
 
+    /**
+     * In Kotlin all expressions can be annotated with annotations with the corresponding annotation target.
+     */
+    @Getter
+    @SuppressWarnings("unchecked")
+    @FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
+    @EqualsAndHashCode(callSuper = false, onlyExplicitlyIncluded = true)
+    final class AnnotatedExpression implements K, Expression {
+
+        @With
+        UUID id;
+
+        @With
+        Space prefix;
+
+        @With
+        Markers markers;
+
+        @With
+        List<J.Annotation> annotations;
+
+        @With
+        Expression expression;
+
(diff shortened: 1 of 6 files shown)


knutwannheden:
@traceyyoshima Adding you as reviewer, since I am unsure if I maybe missed something when adding the new LST type.

src/main/java/org/openrewrite/kotlin/tree/K.java
```diff
@@ -320,6 +320,66 @@ public <S> S service(Class<S> service) {
...
+    @SuppressWarnings("unchecked")
+    @FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
+    @EqualsAndHashCode(callSuper = false, onlyExplicitlyIncluded = true)
+    final class AnnotatedExpression implements K, Expression {
+
+        @With
+        UUID id;
+
+        @With
+        Space prefix;
```
knutwannheden, line 336: I suppose the prefix could instead be pushed down to the first annotation, if we don't want it here.

traceyyoshima: Yeah, I think the prefix should be on the annotation to prevent as many unintentional changes from J as possible.

`ExpressionStatement`, `StatementExpression`, etc. work in a similar way. The `K` tree is just a pass-through to J.

traceyyoshima (changes requested):
I agree with your observation about the prefix. I think it's ideal for the prefix to exist in the same place as it would from the Java compiler.

src/main/java/org/openrewrite/kotlin/tree/K.java
knutwannheden: OK. Thanks for the feedback. I will move that prefix then.

Issue #266, opened 2023-08-30
Name of function type parameters is not parsed


knutwannheden:
[Function types](https://kotlinlang.org/docs/lambdas.html#function-types) can in Kotlin optionally specify parameter names:

```java
    @Test
    void namedParameter() {
        rewriteRun(
          kotlin(
            """
              val f: (p: Any) -> Boolean = { p -> true }
              """
          )
        );
    }
```
The parser does not yet support this and it results in a parse-to-print failure when encountered.

traceyyoshima:
I think `K.FunctionType.Parameters` and removing the `J.Lambda` is reasonable. It'll likely prevent unexpected recipe changes from `J` as well.

Issue #275, opened 2023-08-31
Parse error on nullable parenthesized function type


knutwannheden:
When parsing the [function type](https://kotlinlang.org/docs/lambdas.html#function-types) in
```kt
val v: ((Any) -> Any)? = null
```
what gets printed back out again is
```kt
val v: (Any)(Any) -> Any)? = null
```
It seems like the parser can't handle the parentheses which are there to make the type nullable.

traceyyoshima:
FYI - The compiler removes Unnecessary parens, so the detection is attempted in `visitElement`. However, before the compile revision to access the PSI, there needed to be more information on the list parens. I handled a handful of cases manually and intended to remove all of that code, and I may have missed a few cases.

knutwannheden:
As I think we've seen before, we need something like a `K.FunctionType` as this isn't really a lambda.

PR #282, opened 2023-09-04
Named function type parameters support

knutwannheden:
Up until now the function type were in the LST represented using lambdas. This approach falls down, because with function type parameters the type is mandatory and the name is optional, whereas for lambdas it is the other way around. Also, a lambda as a body, whereas a function type has a return type.

There is now a new `K.FunctionType.Parameter` type used by `K.FunctionType#parameters` and also the fields `Space arrow` for the arrow prefix and `TypedTree returnType` for the return type.

Fixes: #266


knutwannheden:
This is still very much work in progress.

knutwannheden:
@traceyyoshima I need to split this PR in two. I started working on the function type parameters and then noticed that there are more cases when the constructor delegation needs fixing.

The code is still a bit messy (which is also partly due to the fact that our parser currently is designed to advance serially along the contents of the source file rather than extract information from the PSI), but the idea is that when we have constructor delegation (as in e.g. `class Test : Super()` or in `class Test(val id2 : Int) : Super(1 + 3)`) we don't add a `K.FunctionType` (with a `J.NewClass` as its "body") to `J.Class#implementings`, but rather a regular `J.Identifier` (in the example that would be `Super`) and at the same time we add a constructor (`J.MethodDeclaration`) with a `super` call (`J.MethodInvocation`) to its body.

I think this is more correct with respect to what the compiler backend does. But it will continue to be brittle (i.e. Java recipes could easily wreak havoc when modifying Kotlin LSTs).

knutwannheden:
@traceyyoshima What are your thoughts on this? Do you think it is overkill to mimic the Kotlin compiler by inferring synthetic elements like a constructor method declaration and with a body with a `super()` or `this()` call in it and then use that in the printer? There would certainly be other alternatives (e.g. attach more information directly to the marker).

I think for Java interoperability and existing find recipes it may make sense to go the extra mile here.

I will now try to make sure we can maintain deserialization compatibility with the old LSTs.

knutwannheden:
I've now extracted a part of this PR into #283. Will remove the corresponding part from here, once that is merged.

src/main/java/org/openrewrite/kotlin/tree/K.java
```diff
@@ -708,30 +725,138 @@ public List<Modifier> getModifiers() {
...
         @Nullable
+        @With
+        @Getter
         JRightPadded<NameTree> receiver;
 
-        @Override
-        public @Nullable JavaType getType() {
-            return typedTree.getType();
+        @Nullable
+        JContainer<TypeTree> parameters;
```
knutwannheden, line 733: Note: This is `JContainer<TypeTree>` rather than `JContainer<Parameter>` to allow for `J.Empty` when empty.

knutwannheden:
@traceyyoshima This PR is now ready for review. Please note that it also contains basic support for nullable function types (see #275). That is not complete yet, because it will require adding a few more `Space` properties to the LST. But I think we can deal with that later (I will just keep the issue open for now).

Please pay extra attention when reviewing the LST changes. I still struggle a bit with finding the best way to store whitespace in the LST (using `Space`, `JLeftPadded`, `JRightPadded`, and `JContainer` and the corresponding Kotlin types). Possibly that can still be improved / simplified.

knutwannheden:
@traceyyoshima Also here I went ahead and merged, so that I can continue working on other bugs / features without having to do too much conflict resolution in Git later. I hope we will only have to adjust details in this PR.

Issue #233, opened 2023-08-17
Parser maps indexed access operator to `K.Binary`


knutwannheden:
Currently the parser maps the indexed access operator to `K.Binary`. So the example

```kt
val arr = IntArray ( 1 )
val a = arr [ 0 ]
```

is translated to a `K.Binary` where the left is the identifier `a` and the right is the literal `0`.

I suspect it would make more sense to translate this [indexed access operator](https://kotlinlang.org/docs/operator-overloading.html#indexed-access-operator) to a method invocation (in this case to `kotlin.IntArray#get()`) which is the designated `get` operator or possibly something which would be similar to a Java array access.

For more complex examples with multiple parameters it should probably be a method invocation. E.g.

```kt
class Surface {
    operator fun get(x: Int, y: Int) = 2 * x + 4 * y - 10
}
fun getZValue(mySurface: Surface, xValue: Int, yValue: Int) =
    mySurface[xValue, yValue]
```

traceyyoshima:
> @traceyyoshima What are your thoughts on this one?

@knutwannheden I agree entirely. I planned to add this in the `KBinary` changes before I got caught up in COBOL. This will be especially important for preserving the overloads in the LST.

The idea was to add `KArrayAccess,` `KUnary,` `KBinary,` or something along those lines for print idempotence.

knutwannheden:
@traceyyoshima Something like `KArrayAccess` (or `KIndexedArrayAccess`) would work. Another potentially simpler alternative would be to map it to a `J.MethodInvocation` and add an `IndexedAccess` marker to it. I think that could have some nice advantages. In the following example
```kt
class Surface {
    operator fun get(x: Int, y: Int) = 2 * x + 4 * y - 10
}
val surface = Surface()
val z1 = surface[4, 2]
val z2 = surface.get(4, 2)
```
The initializers for `z1` and `z2` would map to the LST (except for the marker of course).

I will prepare a corresponding PR.

PR #288, opened 2023-09-07
Map indexed access `get` operator to method invocation

knutwannheden:
While the Kotlin index access `get` operator is an operator, its syntax is more like a regular method invocation, with the difference that the method name (`get`) is implicit and that the arguments appear inside square brackets rather than regular parentheses. Also, under the hood, this will of course get mapped to a method invocation (which is also mostly true for Kotlin's other operators).

This commit changes the parsing to map the `get` operator to a `J.MethodInvocation` rather than a `K.Binary` and adds an `IndexedAccess` marker to it, so that it can get printed correctly again.

Fixes: #233


knutwannheden:
@traceyyoshima I felt pretty confident about this one, so I went ahead and merged it. We can always revert / fix later. I hope you are fine with that!

PR #290, opened 2023-09-08
Refactor constructor delegation

knutwannheden:
Add new `K.ConstructorInvocation` type and `K.Constructor` to be used for constructor delegation when specified in class header or on any other constructor.


knutwannheden:
@traceyyoshima I think this can be reviewed now.

src/main/java/org/openrewrite/kotlin/KotlinIsoVisitor.java
```diff
@@ -32,6 +32,16 @@ public K.Binary visitBinary(K.Binary binary, P p) {
         return (K.Binary) super.visitBinary(binary, p);
     }
 
+    @Override
+    public K.Constructor visitConstructor(K.Constructor constructor, P p) {
+        return (K.Constructor) super.visitConstructor(constructor, p);
+    }
+
+    @Override
+    public K.ConstructorInvocation visitConstructorInvocation(K.ConstructorInvocation destructuringDeclaration, P p) {
```
traceyyoshima, line 41:
```suggestion
    public K.ConstructorInvocation visitConstructorInvocation(K.ConstructorInvocation constructorInvocation, P p) {
```

src/main/java/org/openrewrite/kotlin/KotlinIsoVisitor.java
```diff
@@ -32,6 +32,16 @@ public K.Binary visitBinary(K.Binary binary, P p) {
...
+        return (K.ConstructorInvocation) super.visitConstructorInvocation(destructuringDeclaration, p);
```
traceyyoshima, line 42:
```suggestion
        return (K.ConstructorInvocation) super.visitConstructorInvocation(constructorInvocation, p);
```

traceyyoshima (approved):
Looks good! I'm curious about what your thoughts are on whether we should be adding SignatureBuilding and TypeMapping tests as the parsing is improved?

kunli2 (approved):

PR #309, opened 2023-09-12
PSI based parser

knutwannheden:
Proof-of-concept for PSI-based parsing.


traceyyoshima (approved):

src/main/java/org/openrewrite/kotlin/tree/K.java
```diff
@@ -505,6 +505,10 @@ public CoordinateBuilder.Expression getCoordinates() {
         }
 
         public enum Type {
+            Plus,
+            Minus,
+            Div,
+            Mul,
```
knutwannheden, line 511: Our parser doesn't appear to be using these four enum constants anywhere. I will remove them for now, as we already have corresponding constants in `J.Binary.Type` (`Addition`, `Subtraction`, `Multiplication`, and ,`Division`).

PR #311, opened 2023-09-12
Fix type attribution for extension methods

knutwannheden:
Extension method calls should be type attributed with a `JavaType.Method` where the receiver type of the extension method is the first parameter.

### DIFF
--- a/src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
+++ b/src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
@@ -722,14 +722,18 @@ class KotlinTypeMapping(typeCache: JavaTypeCache, firSession: FirSession) : Java
             simpleFunction = symbol.fir as FirSimpleFunction
         }
         var paramNames: MutableList<String>? = null
+        if (simpleFunction?.receiverParameter != null) {
+            paramNames = ArrayList(simpleFunction.valueParameters.size + 1)
+            paramNames.add("receiver")
+        }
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
```diff
@@ -722,14 +722,18 @@ class KotlinTypeMapping(typeCache: JavaTypeCache, firSession: FirSession) : Java
             simpleFunction = symbol.fir as FirSimpleFunction
         }
         var paramNames: MutableList<String>? = null
+        if (simpleFunction?.receiverParameter != null) {
+            paramNames = ArrayList(simpleFunction.valueParameters.size + 1)
+            paramNames.add("receiver")
```
knutwannheden, line 727: I need to check what the actual name of this parameter is. Does Kotlin generate it into the bytecode? Is it fixed?

sambsnyd (approved):

src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
sambsnyd: Fastest way to answer this question is probably to hit a breakpoint inside an extension function and see what intelliJ calls it

knutwannheden: I decompiled a class file and it looks like the parameter name is `$this$<extension-name>`. I left it at `$this$` for now.

Issue #315, opened 2023-09-18
Anonymous object  print idempotent issue


kunli2:
Found this bug print idempotent issue during psi-based parser developing

```java
    @Test
    void anonymousObject() {
        rewriteRun(
          kotlin(
            """
              open class Test
              val o  : Test   =  object   :     Test  (   )     {    }"
              """
          )
        );
    }
```

error:
```diff
org.opentest4j.AssertionFailedError: [Bug in source parser or printer resulted in the following difference for "openRewriteFile.kt":
diff --git a/openRewriteFile.kt b/openRewriteFile.kt
index 3d66e08..19a7502 100644
--- a/openRewriteFile.kt
+++ b/openRewriteFile.kt
@@ -1,2 +1,2 @@ 
 open class Test
-val o  : Test   =  object   :     Test  (   )     {    }"
\ No newline at end of file
+val o  : Test   =  object     :   Test  (   )     {    }"
\ No newline at end of file
] 
expected: 
  "open class Test
  val o  : Test   =  object   :     Test  (   )     {    }""
 but was: 
  "open class Test
  val o  : Test   =  object     :   Test  (   )     {    }""
```

PR #316, opened 2023-09-19
Fix anonymous object print idempotent issue

kunli2:
This PR is intended to fix the issue https://github.com/openrewrite/rewrite-kotlin/issues/315

I think the issue is caused by unreasonable space arragement.
For instance, code
```java
open class Test
val o  : Test   =  object   :     Test  (   )     {    }
```

identify the space before and after `object`
```java
val o  : Test   = /*space1*/ object /*space2*/  :  /*space3*/   Test  (   )     {    }
```

Before:
`space1` is the prefix of `object` in a marker `KObject`.
`space2` is the prefix of `new`  (or TypeExpression)
`space3` is the prefix of the entire initializer (new class)

Proposed in this Fix:
`space1` is prefix of the entire initializer (new class)
`space2` is the prefix of `TypeReferencePrefix` marker.
`space3` is the prefix of `new`  (or TypeExpression)

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/marker/KObject.java
+++ b/src/main/java/org/openrewrite/kotlin/marker/KObject.java
@@ -26,10 +26,10 @@ import java.util.UUID;
 @With
 public class KObject implements Marker {
     UUID id;
-    Space prefix;
+    Space suffix;
 
-    public KObject(UUID id, Space prefix) {
+    public KObject(UUID id, Space suffix) {
         this.id = id;
-        this.prefix = prefix;
+        this.suffix = suffix;
     }
 }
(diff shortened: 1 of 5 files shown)


src/main/java/org/openrewrite/kotlin/marker/KObject.java
```diff
@@ -26,10 +26,10 @@
 @With
 public class KObject implements Marker {
     UUID id;
-    Space prefix;
+    Space suffix;
 
-    public KObject(UUID id, Space prefix) {
+    public KObject(UUID id, Space suffix) {
```
knutwannheden, line 31: We have to be careful here. I am not 100% sure what this means for LST backward-compatibility. AFAIK Jackson should use the constructor, but I don't know if it would complain because the parameter name doesn't align with the serialized property name (`prefix` vs. `suffix`). I almost think you would have to test it.

Issue #318, opened 2023-09-19
Extension property getter/setter parse-to-print idempotent issue 


kunli2:
Found this extension property getter parse-to-print idempotent issue 

```java
    @Test
    void inline() {
        rewriteRun(
          kotlin(
            """
              class Spec
              inline val  Spec   .    `java-base`     : String  get   (    )   =  "  "
              """
          )
        );
    }
```
Error
```diff

org.opentest4j.AssertionFailedError: [Bug in source parser or printer resulted in the following difference for "openRewriteFile.kt":
diff --git a/openRewriteFile.kt b/openRewriteFile.kt
index 7de773f..c8d7a1d 100644
--- a/openRewriteFile.kt
+++ b/openRewriteFile.kt
@@ -1,2 +1,2 @@ 
 class Spec
-inline val  Spec   .    `java-base`     : String  get   (    )   =  "  "
\ No newline at end of file
+inline val  Spec   .    `java-base`     : String  get   (   )   =  "  "
\ No newline at end of file
] 
expected: 
  "class Spec
  inline val  Spec   .    `java-base`     : String  get   (    )   =  "  ""
 but was: 
  "class Spec
  inline val  Spec   .    `java-base`     : String  get   (   )   =  "  ""
```

PR #319, opened 2023-09-19
Fix extension property getter/setter parse-to-print idempotent issue

kunli2:
fixes https://github.com/openrewrite/rewrite-kotlin/issues/318

I found this is not an easy space-related fix, but need to change the model.
The root cause is that `getter`/`setter` is supposed to have a `J.Empty` as a parameter, however, it's overridden to store the receiver name, and accordingly, the stored space is erased.
To fix this problem, introduced a new field `receiverName` to `K.Property` to store the receiver name.

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/tree/K.java
+++ b/src/main/java/org/openrewrite/kotlin/tree/K.java
@@ -1418,7 +1418,17 @@ public interface K extends J {
 
         boolean isSetterFirst;
 
-        public Property(UUID id, Space prefix, Markers markers, @Nullable JContainer<TypeParameter> typeParameters, VariableDeclarations variableDeclarations, @Nullable J.MethodDeclaration getter, @Nullable J.MethodDeclaration setter, boolean isSetterFirst) {
+        @Nullable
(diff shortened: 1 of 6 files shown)


src/main/java/org/openrewrite/kotlin/tree/K.java
```diff
@@ -1418,7 +1418,17 @@ public List<TypeParameter> getTypeParameters() {
 
         boolean isSetterFirst;
 
-        public Property(UUID id, Space prefix, Markers markers, @Nullable JContainer<TypeParameter> typeParameters, VariableDeclarations variableDeclarations, @Nullable J.MethodDeclaration getter, @Nullable J.MethodDeclaration setter, boolean isSetterFirst) {
+        @Nullable
+        JRightPadded<J> receiverName;
+
+        @Nullable
+        public J getReceiverName() {
```
knutwannheden, line 1425: I suggest we ho with `getReceiver()` instead. Since this is a new property I think it would have been nice with a more specific type. Although `J.Identifier` and `J.FieldAccess` may unfortunately not have anything in common except for `J` and `Expression`.

Issue #326, opened 2023-09-22
Parsing error on function call


knutwannheden:
The test
```java
    @Test
    void functionWithFunctionTypeParameter() {
        rewriteRun(
          kotlin(
            """
              class GradleSpigotDependencyLoaderTestBuilder(
                  var init: TestInitializer.() -> Unit = {}
              ) {
                  fun init(builder: TestInitializer.() -> Unit) {
                      init = builder
                  }
              }
              
              class TestInitializer(
                  val resourcesDir: String
              )
              
              fun runTest() {
                  val builder = GradleSpigotDependencyLoaderTestBuilder()
                  builder.init(TestInitializer("null"))
              }
              """
          )
        );
    }
```
fails with a parse to print failure:
```diff
diff --git a/openRewriteFile.kt b/openRewriteFile.kt
index 1cbbb68..900e864 100644
--- a/openRewriteFile.kt
+++ b/openRewriteFile.kt
@@ -12,5 +12,5 @@ 
 
 fun runTest() {
     val builder = GradleSpigotDependencyLoaderTestBuilder()
-    builder.init(TestInitializer("null"))
+    TestInitializerbuilde.initr.init(TestInitializer("null")
 }
\ No newline at end of file
```

PR #327, opened 2023-09-27
Fix a parsing error on function call

kunli2:
Attempt to fix https://github.com/openrewrite/rewrite-kotlin/issues/326
the reason is when parsing a function call, the select is wrongly set to a parameter due to some wrong conditions.
the root cause is a function call like `builder.init(TestInitializer("null"))` should be an explicit function call, however, it's a `FirImplicitInvokeCall`.

### DIFF
--- a/src/main/kotlin/org/openrewrite/kotlin/internal/KotlinParserVisitor.kt
+++ b/src/main/kotlin/org/openrewrite/kotlin/internal/KotlinParserVisitor.kt
@@ -1207,9 +1207,9 @@ class KotlinParserVisitor(
             }
 
             if (functionCall is FirImplicitInvokeCall) {
-                val dispatchReceiver = functionCall.dispatchReceiver
-                if (dispatchReceiver is FirPropertyAccessExpression && dispatchReceiver.explicitReceiver != null) {
-                    select = padRight(convertToExpression<Expression>(dispatchReceiver.explicitReceiver as FirElement, data)!!, whitespace())
+                val receiver = functionCall.explicitReceiver
+                if (receiver is FirPropertyAccessExpression && receiver.explicitReceiver != null) {
+                    select = padRight(convertToExpression(receiver.explicitReceiver as FirElement, data)!!, whitespace())
                 }
             }
 
--- a/src/main/kotlin/org/openrewrite/kotlin/internal/KotlinParserVisitor.kt
+++ b/src/main/kotlin/org/openrewrite/kotlin/internal/KotlinParserVisitor.kt
@@ -1502,7 +1502,7 @@ class KotlinParserVisitor(
                 val expression = flattenedExpressions[i]
 
                 // Didn't find a way to proper reset the cursor, so have to do a hard reset here
-                if (hasParentheses) {
+                if (hasParentheses && i == 0) {
                     cursor = lPAROffset + 1
                 }
 
--- a/src/main/kotlin/org/openrewrite/kotlin/internal/KotlinParserVisitor.kt
+++ b/src/main/kotlin/org/openrewrite/kotlin/internal/KotlinParserVisitor.kt
@@ -1177,9 +1177,9 @@ class KotlinParserVisitor(
 
             var hasExplicitReceiver = false
             if (functionCall is FirImplicitInvokeCall) {
-                val dispatchReceiver = functionCall.dispatchReceiver
-                if (dispatchReceiver is FirPropertyAccessExpression) {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/openrewrite/kotlin/internal/KotlinParserVisitor.kt
```diff
@@ -1175,10 +1175,22 @@ class KotlinParserVisitor(
                 markers = markers.addIfAbsent(Extension(randomId()))
             }
 
-            val implicitExtensionFunction = functionCall is FirImplicitInvokeCall
+            var hasExplicitReceiver = false
+            if (functionCall is FirImplicitInvokeCall) {
+                val dispatchReceiver = functionCall.dispatchReceiver
```
traceyyoshima, line 1180: Typically, the compiler uses the explicit receiver if it's not null then uses the dispatch receiver otherwise.

src/main/kotlin/org/openrewrite/kotlin/internal/KotlinParserVisitor.kt
```diff
@@ -1194,6 +1206,13 @@ class KotlinParserVisitor(
                 }
             }
 
+            if (functionCall is FirImplicitInvokeCall) {
+                val dispatchReceiver = functionCall.dispatchReceiver
+                if (dispatchReceiver is FirPropertyAccessExpression && dispatchReceiver.explicitReceiver != null) {
```
traceyyoshima, line 1211: Similar concept here to explicitReceiver

src/main/kotlin/org/openrewrite/kotlin/internal/KotlinParserVisitor.kt
```diff
@@ -1413,6 +1432,18 @@ class KotlinParserVisitor(
         var callPsi = getPsiElement(firCall)!!
         callPsi = if (callPsi is KtDotQualifiedExpression || callPsi is KtSafeQualifiedExpression) callPsi.lastChild else callPsi
         val firArguments = if (skipFirstArgument) firCall.argumentList.arguments.subList(1, firCall.argumentList.arguments.size) else firCall.argumentList.arguments
+
+        var hasParentheses = false
+        var lPAROffset = 0
+        if (firCall.argumentList.source is KtRealPsiSourceElement) {
+            val firArgumentsSource = firCall.argumentList.source.psi
+            val firstChild = firArgumentsSource?.firstChild
+            if (firstChild != null && firstChild.node.elementType.equals(KtTokens.LPAR)) {
```
traceyyoshima, line 1441: Replace equals with either == or KtTokens.LPAR.equals(...)

src/main/kotlin/org/openrewrite/kotlin/internal/KotlinParserVisitor.kt
```diff
@@ -1469,6 +1500,12 @@ class KotlinParserVisitor(
             for (i in flattenedExpressions.indices) {
                 isTrailingLambda = hasTrailingLambda && i == argumentCount - 1
                 val expression = flattenedExpressions[i]
+
+                // Didn't find a way to proper reset the cursor, so have to do a hard reset here
+                if (firCall is FirImplicitInvokeCall && hasParentheses && i == 0) {
```
traceyyoshima, line 1505: Can you share info on why this is necessary?

src/main/kotlin/org/openrewrite/kotlin/internal/KotlinParserVisitor.kt
kunli2: For the giving test case.
```
builder.init(TestInitializer("null"))
```
it has both an explicit receiver and a dispatch receiver.
however, the explicit receiver is `init` but not `builder`. but the dispatch receiver is the `builder`.

traceyyoshima: I checked line `1180`, and the values are equal. The explicit receiver and the dispatch receiver are the same object. The compiler typically uses the explicit receiver when it is not null and falls back to the dispatch receiver otherwise. The code only uses the dispatch receiver, so I'm trying to understand what you mean.

kunli2: Yeah, the explicit receiver and dispatch receiver are equal here, revised to use the explicit receiver.

traceyyoshima (approved):

PR #330, opened 2023-09-28
Add new Kotlin formatting options

knutwannheden:
`WrappingAndBraces` now has `ExpressionBodyFunctions(boolean useContinuationIndent)` and `ElvisExpressions(boolean useContinuationIndent)`


knutwannheden:
@kunli2 I believe there is still quite a bit of the formatting which still isn't the same as in IntelliJ, but I think it is already a bit closer now, since by default the IntelliJ formatting in many places doesn't use the continuation indents, but rather just regular indents and I believe this is what we've seen a lot in the SaaS as well.

kunli2 (approved):

PR #331, opened 2023-09-29
To support seperate alias import

kunli2:
To support Intellij's Kotlin import layout import aliases separately feature,
Screenshot 2023-09-29 at 11 51 25 AM

This setting is by default enabled on the UI but in reality, it's off by default (no change if you auto-format your code) in the IDE. so we set it to false as default, so by default, when adding an alias import, it will be added at the proper position.

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/format/AutoFormatVisitor.java
+++ b/src/main/java/org/openrewrite/kotlin/format/AutoFormatVisitor.java
@@ -80,6 +80,7 @@ public class AutoFormatVisitor<P> extends KotlinIsoVisitor<P> {
 
         t = new RemoveTrailingWhitespaceVisitor<>(stopAfter).visit(t, p, cursor.fork());
 
+        t = new ImportReorderingVisitor<>().visit(t, p, cursor.fork());
         return t;
     }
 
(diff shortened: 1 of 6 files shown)


knutwannheden:
@kunli2 With this option enabled, if I run auto-format on:
```kt
import java.util.regex.Pattern as Pat
import java.util.*

class T(s: Scanner, p: Pat)
```
I end up with:
```kt
import java.util.*
import java.util.regex.Pattern as Pat

class T(s: Scanner, p: Pat)
```
So AFAICT the option does have an effect.

knutwannheden:
(With the option disabled, the auto-formatter does not change the import order.)

kunli2:
> @kunli2 With this option enabled, if I run auto-format on:
> 
> ```kotlin
> import java.util.regex.Pattern as Pat
> import java.util.*
> 
> class T(s: Scanner, p: Pat)
> ```
> 
> I end up with:
> 
> ```kotlin
> import java.util.*
> import java.util.regex.Pattern as Pat
> 
> class T(s: Scanner, p: Pat)
> ```
> 
> So AFAICT the option does have an effect.

it's because we don't have import-reordering-visitor in the autoFormat, seems we need one, I added one in this PR.

PR #338, opened 2023-10-10
Fixed type attribution for cone captured types.

traceyyoshima:
Changes:
- ConeCaptureTypes of generic types: `*` will produce the correct type signature and type.
- Added a test to show that the dispatch receiver on the field access contains the correct bounds.

### DIFF
--- a/src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
+++ b/src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
@@ -21,6 +21,7 @@ import org.jetbrains.kotlin.descriptors.ClassKind
 import org.jetbrains.kotlin.descriptors.Modality
 import org.jetbrains.kotlin.descriptors.Visibility
 import org.jetbrains.kotlin.fir.FirSession
+import org.jetbrains.kotlin.fir.analysis.checkers.getContainingClassSymbol
 import org.jetbrains.kotlin.fir.containingClassLookupTag
 import org.jetbrains.kotlin.fir.declarations.*
 import org.jetbrains.kotlin.fir.declarations.utils.isLocal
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
```diff
@@ -996,8 +999,12 @@ class KotlinTypeMapping(typeCache: JavaTypeCache, firSession: FirSession) : Java
         val isGeneric = type is ConeKotlinTypeProjectionIn ||
                 type is ConeKotlinTypeProjectionOut ||
                 type is ConeStarProjection ||
-                type is ConeTypeParameterType
+                type is ConeTypeParameterType ||
+                type is ConeCapturedType
         if (isGeneric) {
+            if (type is ConeCapturedType) {
+                println()
```
knutwannheden, line 1006: Some remnant debug code?

src/main/kotlin/org/openrewrite/kotlin/KotlinTypeSignatureBuilder.kt
```diff
@@ -345,11 +345,6 @@ class KotlinTypeSignatureBuilder(private val firSession: FirSession) : JavaTypeS
             s.append("Generic{in ")
             s.append(signature(type1))
             s.append("}")
-        } else if (type is ConeCapturedType) {
```
traceyyoshima, line 348: Hi @knutwannheden, I'd like an extra pair of eyes here if I overlook why this was added.
AFAICT, the signature(type1) on line 351 always returns `Generic{in {undefined}}` and the ConeCaptureType was covered on line `392`.

src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
traceyyoshima: Ah yes, I removed that, haha -- ty

src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
```diff
@@ -878,11 +879,13 @@ class KotlinTypeMapping(typeCache: JavaTypeCache, firSession: FirSession) : Java
...
         var resolvedOwner: JavaType? = owner
         if (owner == null) {
-            val lookupTag: ConeClassLikeLookupTag? = symbol.containingClassLookupTag()
-            if (lookupTag != null && lookupTag.toSymbol(firSession) !is FirAnonymousObjectSymbol) {
-                // TODO check type attribution for `FirAnonymousObjectSymbol` case
-                resolvedOwner =
-                    type(lookupTag.toFirRegularClassSymbol(firSession)!!.fir) as JavaType.FullyQualified?
+            // TODO: fix type attribution for anonymous objects. Currently, returns JavaType.Unknown.
+            if (symbol.dispatchReceiverType != null) {
+                resolvedOwner = type(symbol.dispatchReceiverType)
```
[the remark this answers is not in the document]
knutwannheden: Just for context: What I did was to move these lines from the parser here, so that we don't need to duplicate this logic for the PSI-based parser: https://github.com/openrewrite/rewrite-kotlin/commit/6e46b0ea00990f1c8c90f05df951aefa399eb1cf#diff-0f192ceb739be61c09f94d0991022e59b90560191b2760a9b297692ccaac212dL4748-L4754

knutwannheden (approved):
👍 LGTM!

src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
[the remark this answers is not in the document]
traceyyoshima: Ahhh, that makes sense -- tysm!

PR #339, opened 2023-10-10
top-level function declaration type fixes.

traceyyoshima:
Changes:
- Function signature and types will contain the correct type from a ConeClassLikeType instead of JavaType.Unknown.
- Added FirFileSymbol as field to SignatureBuilder and TypeMapping to cover top-level declarations.
- Fixed signature and type for the owner of JavaType$Variable from top-level method declaration method parameters.
- Kotlin generic bounds in and out {type} will contain the equivalent java name `?`
- ConeIntersectionType will return a generic with the correct bounds instead of JavaType.Unknown.

### DIFF
--- a/src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
+++ b/src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
@@ -58,15 +58,17 @@ import org.openrewrite.kotlin.KotlinTypeSignatureBuilder.Companion.convertFileNa
 import org.openrewrite.kotlin.KotlinTypeSignatureBuilder.Companion.convertKotlinFqToJavaFq
 
 @Incubating(since = "0.0")
-class KotlinTypeMapping(typeCache: JavaTypeCache, firSession: FirSession) : JavaTypeMapping<Any> {
+class KotlinTypeMapping(typeCache: JavaTypeCache, firSession: FirSession, firFileSymbol: FirFileSymbol) : JavaTypeMapping<Any> {
     private val signatureBuilder: KotlinTypeSignatureBuilder
     private val typeCache: JavaTypeCache
     private val firSession: FirSession
+    private val firFileSymbol: FirFileSymbol
 
     init {
-        signatureBuilder = KotlinTypeSignatureBuilder(firSession)
+        signatureBuilder = KotlinTypeSignatureBuilder(firSession, firFileSymbol)
         this.typeCache = typeCache
         this.firSession = firSession
+        this.firFileSymbol = firFileSymbol
     }
 
     override fun type(type: Any?): JavaType {
@@ -184,6 +186,8 @@ class KotlinTypeMapping(typeCache: JavaTypeCache, firSession: FirSession) : Java
                     if (classifierSymbol != null && classifierSymbol.fir is FirTypeParameter) {
                         return resolveConeTypeProjection(classifierSymbol.fir as FirTypeParameter, signature)
                     }
+                } else if (coneKotlinType is ConeClassLikeType) {
+                    return type(coneKotlinType, ownerFallBack)
                 }
                 return classType(type, signature, ownerFallBack)
             }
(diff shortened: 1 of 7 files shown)


src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
```diff
@@ -58,15 +58,17 @@ import org.openrewrite.kotlin.KotlinTypeSignatureBuilder.Companion.convertFileNa
 import org.openrewrite.kotlin.KotlinTypeSignatureBuilder.Companion.convertKotlinFqToJavaFq
 
 @Incubating(since = "0.0")
-class KotlinTypeMapping(typeCache: JavaTypeCache, firSession: FirSession) : JavaTypeMapping<Any> {
+class KotlinTypeMapping(typeCache: JavaTypeCache, firSession: FirSession, firFileSymbol: FirFileSymbol) : JavaTypeMapping<Any> {
```
traceyyoshima, line 61: The ownerStack will be wrong without access to the file when a `SimpleFunction` is on the top of the stack for the parameters in a top-level method declaration.

src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
```diff
@@ -546,7 +550,7 @@ class KotlinTypeMapping(typeCache: JavaTypeCache, firSession: FirSession) : Java
                 }
             }
             if (resolvedDeclaringType == null) {
-                return null
+                resolvedDeclaringType = TypeUtils.asFullyQualified(type(firFileSymbol.fir))
```
traceyyoshima, line 553: @knutwannheden I needed to figure out where we should fall back to the FirFileSymbol. This "could" technically set the incorrect owner if we don't assign the owner appropriately, but it also isolates incorrect types to each file.

Curious, what do you think about falling back onto the FirFileSymbol by default?

src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
```diff
@@ -882,6 +886,8 @@ class KotlinTypeMapping(typeCache: JavaTypeCache, firSession: FirSession) : Java
             // TODO: fix type attribution for anonymous objects. Currently, returns JavaType.Unknown.
             if (symbol.dispatchReceiverType != null) {
                 resolvedOwner = type(symbol.dispatchReceiverType)
+            } else if (ownerFallBack is FirFunctionSymbol<*>) {
+                resolvedOwner = methodDeclarationType(ownerFallBack.fir, null, firFileSymbol)
```
traceyyoshima, line 890: The signature is set correctly to the method declaration signature. This change fixes the JavaType returned by the signature.

src/test/java/org/openrewrite/kotlin/KotlinTypeMappingTest.java
```diff
@@ -365,6 +369,32 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Atomi
...
+                  val l = listOf ( "foo" to "1" , "bar" to 2 )
+                  """, spec -> spec.afterRecipe(cu -> {
+                    MethodMatcher methodMatcher = new MethodMatcher("kotlin.collections.CollectionsKt listOf(..)");
+                    AtomicBoolean found = new AtomicBoolean(false);
+                    new KotlinIsoVisitor<AtomicBoolean>() {
+                        @Override
+                        public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, AtomicBoolean atomicBoolean) {
+                            if (methodMatcher.matches(method)) {
+                                assertThat(method.getMethodType().toString())
+                                  .isEqualTo("kotlin.collections.CollectionsKt{name=listOf,return=kotlin.collections.List<kotlin.Pair<kotlin.String, Generic{kotlin.Comparable<Generic{*}> & java.io.Serializable}>>,parameters=[kotlin.Array<Generic{? extends Generic{T}}>]}");
```
traceyyoshima, line 386: Just so you know -- In Java, a generic type intersection contains the bounded generic name like T, V, U. So the type becomes `Generic{<name> extends <typeA> & <typeB>}`.

In Kotlin, the ConeIntersectionType contains a generic name `it` and appears as `it(<typeA>, <typeB>)`. So, I left the name out.

src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
knutwannheden: > Curious, what do you think about falling back onto the FirFileSymbol by default?

I think that sounds reasonable. I wanted to do something similar myself, but first wanted to discuss that with you. So instead I added this owner stack, which is easy enough to remove again.

PR #346, opened 2023-10-17
Present bit operators to J.Binary instead of method invocation

kunli2:
For `!in` binary operator we no longer present it as a method invocation but as a `K.Binary` with `NotContains` type.

Similarly, For bit operators like bit and/or/xor, we should map them to a J.Binary instead of method invocations.


traceyyoshima (approved):
LGTM

PR #347, opened 2023-10-19
IR type mapping and signature building

traceyyoshima:
Changes:
- Types and signatures are mappable through the IR.
- IR-type signature builder tests pass.
- IR-type mapping tests require either testing through the IR or via the LST.
  - I'll add an IR test suite if we'd like one, and we can use the LST through the PSI to match types. Still determining if it's worth it. @kunli2 @knutwannheden, what do you think?

Notes:
- TypeReceivers will be more consistent due to how the symbols are associated with the IR tree.
- Both type mapping and signature building are much more straightforward since types are accessible on the IR tree.
- Parents are available on the tree elements, so it does not need to be inferred.
- inline functions like `println` have a parent class of `kotlin.io`, since the `ConsoleKt` name comes from a `JVM` annotation and is unavailable in the tree.
- generate methods like `equals`, `toString`, etc. are all included in the IR tree.

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/KotlinParser.java
+++ b/src/main/java/org/openrewrite/kotlin/KotlinParser.java
@@ -58,7 +58,6 @@ import org.jetbrains.kotlin.fir.session.FirSessionFactoryHelper;
 import org.jetbrains.kotlin.fir.session.environment.AbstractProjectFileSearchScope;
 import org.jetbrains.kotlin.idea.KotlinFileType;
 import org.jetbrains.kotlin.idea.KotlinLanguage;
-import org.jetbrains.kotlin.ir.declarations.IrFile;
 import org.jetbrains.kotlin.load.kotlin.PackagePartProvider;
 import org.jetbrains.kotlin.modules.Module;
 import org.jetbrains.kotlin.name.Name;
--- a/src/main/java/org/openrewrite/kotlin/KotlinParser.java
+++ b/src/main/java/org/openrewrite/kotlin/KotlinParser.java
@@ -58,7 +58,6 @@ import org.jetbrains.kotlin.fir.session.FirSessionFactoryHelper;
 import org.jetbrains.kotlin.fir.session.environment.AbstractProjectFileSearchScope;
 import org.jetbrains.kotlin.idea.KotlinFileType;
 import org.jetbrains.kotlin.idea.KotlinLanguage;
-import org.jetbrains.kotlin.ir.declarations.IrClass;
 import org.jetbrains.kotlin.ir.declarations.IrFile;
 import org.jetbrains.kotlin.load.kotlin.PackagePartProvider;
 import org.jetbrains.kotlin.modules.Module;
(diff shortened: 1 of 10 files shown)


src/main/java/org/openrewrite/kotlin/KotlinParser.java
```diff
@@ -473,25 +490,38 @@ public CompiledSource parse(List<Parser.Input> sources, Disposable disposable, E
...
+            actualizedResult = convertToIrAndActualizeForJvm(firResult, extensions, irConfiguration, irGenerationExtensions, diagnosticsReporter);
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
 
-        assert kotlinSources.size() == result.getSecond().size();
         List<FirFile> second = result.getSecond();
         for (int i = 0; i < second.size(); i++) {
             kotlinSources.get(i).setFirFile(second.get(i));
         }
```
knutwannheden, line 514: Looks like this loop can be removed.

src/main/java/org/openrewrite/kotlin/KotlinParser.java
```diff
@@ -473,25 +490,38 @@ public CompiledSource parse(List<Parser.Input> sources, Disposable disposable, E
...
+                compilerConfiguration.getBoolean(JVMConfigurationKeys.LINK_VIA_SIGNATURES),
+                compilerConfiguration.putIfAbsent(EVALUATED_CONST_TRACKER, EvaluatedConstTracker.Companion.create())
+        );
+
+        List<IrGenerationExtension> irGenerationExtensions = IrGenerationExtension.Companion.getInstances(projectEnvironment.getProject());
+        Fir2IrActualizedResult actualizedResult;
+        try {
+            actualizedResult = convertToIrAndActualizeForJvm(firResult, extensions, irConfiguration, irGenerationExtensions, diagnosticsReporter);
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
knutwannheden, line 508: Unsure if this is the appropriate way to deal with the exception.

knutwannheden:
From what I can tell there are no drawbacks to this approach, so I think we should proceed. What I am still missing is the use of this in the `KotlinParserVisitor`. How much work do you expect remains to get this to be fully functional?

src/main/java/org/openrewrite/kotlin/KotlinParser.java
traceyyoshima: I'll look into incremental compilation, which will help a lot.

I found that the `convertToIrAndActualizeForJvm` only throws an exception when types cannot be bound. Meaning either a compilation error exists in the project or a test was setup incorrectly.

Now, that we've updated the sources to prevent CRLF, we can add the analysis phase and likely surface compilation errors too

traceyyoshima:
> From what I can tell there are no drawbacks to this approach, so I think we should proceed. What I am still missing is the use of this in the `KotlinParserVisitor`. How much work do you expect remains to get this to be fully functional?

The plan is to add this directly to the PSI-based parser. I ran the type utility method in KotlinIrTypeMapping on the IR files produced by the tests and implemented all the declarations, method calls, variables, properties, etc. So, it's mostly done already.

PR #353, opened 2023-10-30
Add `K.Unary` to present `POSTFIX_EXPRESSION` like `!!`

kunli2:
Use K.Unary to present `POSTFIX_EXPRESSION` like `!!`.

Currently, for the expression like `list [ 0 ] !!`

method invocation `list.get( 0 )` to present it a marker `CheckNotNull` with prefix.
Changed to involve `K.Unary` to present this and will deprecate marker `CheckNotNull`.

And also, `list [ 0 ]` should be a `J.ArrayAccess`.


kunli2:
> consider

Good call, Knut, I will verify those indexed access operators today and add all of them to test, we can discuss it if new Kotlin-specific LST element required

PR #354, opened 2023-11-01
Type mapping revisions.

traceyyoshima:
Changes:
- Parameterized types with generic bounds will retain the bounded type attribution.
    - `Triple<A, B, C> (...)` -> `Triple(1,2,3)` -> `kotlin.Triple<kotlin.Int, kotlin.Int, kotlin.Int>` instead of `kotlin.Triple<Generic{A}, Generic{B}, Generic{C}>`
- TypeMapping:
  - Functions now accept `FirElement`(s) instead of `FirSymbol`(s)
  - Added support for various FirElements, which were not covered.
  - Cache the signatures early to prevent recalculations.
  - Added `createShallowClass` to reuse `JavaType$ShallowClass` types instead of duplication.
  - Filter out `SOURCE` retention annotations on `JavaElemen`t.
  - Updated signatures to match `JavaTypeMapping`.
  - `FileFile` is a default fallback in `TypeMapping`
- PsiElementAssociations: 
  - Fixed most of the PSI to FIR associations.
  - Moved association code from `KotlinTreeParserVisitor` to `PsiElementAssociations`.

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
+++ b/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
@@ -275,16 +275,14 @@ public class KotlinTreeParserVisitor extends KtVisitor<J, ExecutionContext> {
         if (reference != null && reference.getResolvedSymbol() instanceof FirNamedFunctionSymbol) {
             methodReferenceType = psiElementAssociations.getTypeMapping().methodDeclarationType(
                     ((FirNamedFunctionSymbol) reference.getResolvedSymbol()).getFir(),
-                    TypeUtils.asFullyQualified(type(expression.getReceiverExpression())),
-                    owner(expression)
+                    expression.getReceiverExpression()
             );
         }
         JavaType.Variable fieldReferenceType = null;
         if (reference != null && reference.getResolvedSymbol() instanceof FirPropertySymbol) {
             fieldReferenceType = psiElementAssociations.getTypeMapping().variableType(
-                    (FirVariableSymbol<FirVariable>) reference.getResolvedSymbol(),
-                    TypeUtils.asFullyQualified(type(expression.getReceiverExpression())),
-                    owner(expression)
+                    ((FirPropertySymbol) reference.getResolvedSymbol()).getFir(),
+                    expression.getReceiverExpression()
             );
         }
         JRightPadded<Expression> receiver;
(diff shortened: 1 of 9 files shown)


src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
```diff
@@ -2725,7 +2722,7 @@ public J visitProperty(KtProperty property, ExecutionContext data) {
             maybeBeforeSemicolon = prefix(property.getLastChild());
         }
 
-        JavaType.Variable vt = variableType(property);
+        JavaType.Variable vt = variableType(property, owner(property));
```
knutwannheden, line 2725: Since we now only ever use `owner()` like here (either `variableType(psi, owner(psi))` or `type(psi, owner(psi))`), we could also consider doing away with the `ownerStack` and instead just search the PSI tree upwards. Then we would also not need a separate parameter for the owner, but could just do that inside the `PsiElementAssociations` class. There is of course a slight performance overhead with this (always searching the tree upwards), but I think it should be pretty negligible.

I don't have any strong feelings about this. I just wanted to point it out.

knutwannheden (approved):
This PR is very large. I haven't had time to review it all in detail, but I trust your judgement here, @traceyyoshima !

src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
traceyyoshima: I think that's a great point, it should be pretty simple for us to find the parent `KtDeclaration` instead of tracking the owner. For now, I prefer to leave that as a separate change and focus on TypeMapping optimizations.

src/test/java/org/openrewrite/kotlin/KotlinTypeMappingTest.java
```diff
@@ -510,6 +509,47 @@ public K.When visitWhen(K.When when, AtomicBoolean found) {
             );
         }
 
+        @Test
+        void operatorOverload() {
```
traceyyoshima, line 513: This is a little odd -- the FIR associated with a KtBinaryExpression is the method invocation, so the K.BinaryType is a MethodType.

src/test/java/org/openrewrite/kotlin/KotlinTypeMappingTest.java
```diff
@@ -510,6 +512,98 @@ public K.When visitWhen(K.When when, AtomicBoolean found) {
             );
         }
 
+        @ParameterizedTest
+        @CsvSource({
+          "n++,kotlin.Int",
+          "--n,kotlin.Int",
+          "n += a,kotlin.Int",
+          "n = a + b,kotlin.Int{name=plus,return=kotlin.Int,parameters=[kotlin.Int]}"
```
traceyyoshima, line 520: The J.Binary has a method type instead of a boolean. This should probably be detected as a non-overloaded method and assign the return type instead.

PR #355, opened 2023-11-02
Switched to PSI-parser

traceyyoshima:
Changes:

- Removed FIR-based parsing from parsing inputs.
- Removed IR file generation until we can switch to IR-type mapping.
- Removed expected to fail.
- Added `disablePublishingSnapshots` as a failing test to prevent publishing snapshots. The test may be removed if @knutwannheden thinks the changes will be fine with the SAAS.

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
+++ b/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
@@ -25,21 +25,9 @@ import org.jetbrains.kotlin.com.intellij.psi.impl.source.tree.PsiErrorElementImp
 import org.jetbrains.kotlin.com.intellij.psi.tree.IElementType;
 import org.jetbrains.kotlin.com.intellij.psi.util.PsiTreeUtil;
 import org.jetbrains.kotlin.fir.FirElement;
-import org.jetbrains.kotlin.fir.declarations.FirFile;
-import org.jetbrains.kotlin.fir.declarations.FirFunction;
-import org.jetbrains.kotlin.fir.declarations.FirResolvedImport;
-import org.jetbrains.kotlin.fir.declarations.FirVariable;
-import org.jetbrains.kotlin.fir.expressions.FirConstExpression;
-import org.jetbrains.kotlin.fir.expressions.FirFunctionCall;
-import org.jetbrains.kotlin.fir.expressions.FirStringConcatenationCall;
 import org.jetbrains.kotlin.fir.references.FirResolvedCallableReference;
-import org.jetbrains.kotlin.fir.references.FirResolvedNamedReference;
-import org.jetbrains.kotlin.fir.symbols.FirBasedSymbol;
 import org.jetbrains.kotlin.fir.symbols.impl.FirNamedFunctionSymbol;
 import org.jetbrains.kotlin.fir.symbols.impl.FirPropertySymbol;
-import org.jetbrains.kotlin.fir.symbols.impl.FirVariableSymbol;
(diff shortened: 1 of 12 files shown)


src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
```diff
@@ -1833,9 +1821,9 @@ public J visitCallExpression(KtCallExpression expression, ExecutionContext data)
         }
         PsiElementAssociations.ExpressionType type = psiElementAssociations.getCallType(expression);
         if (type == PsiElementAssociations.ExpressionType.CONSTRUCTOR) {
+            JavaType.Method mt = methodInvocationType(expression);
```
traceyyoshima, line 1824: Note:

PSI to FIR association does not guarantee that types are correct via `visitSimpleNameExpression`.
Rather than return `null` by default, I allowed the associated FIR element to set the type. ( which may be a mistake )

So, when applicable, we'll need to compare types from the `JavaParser` to update types on names so that type utilities will work appropriately. Once we notice a pattern, helper methods will help us a lot.

Example:
`val foo = Triple(1, 2, 3)`
The identifier `Triple` is associated with a resolved name reference of a `FirConstructor`.

In Java, a `J.NewClass` has a type of JavaType$Method like `kotlin.Triple<kotlin.Int, kotlin.Int, kotlin.Int>{name=<constructor>,return=kotlin.Triple<kotlin.Int, kotlin.Int, kotlin.Int>,parameters=[kotlin.Int,kotlin.Int,kotlin.Int]}`
and the name would be the class type `kotlin.Triple<kotlin.Int, kotlin.Int, kotlin.Int>` ( I don't know if it would be parameterized.

If the constructor were parameterized, I.E., `ArrayList<String>()`
The `PSI` will likely create a `JavaType$Parameterized` or `JavaType.Method` for the name `ArrayList` due to the associated FIR, and the PSI-parser would create a `J.Parameterized` due to type arguments.

In this case:
```
J.NewClass(
name => J.Parameterized(
                    clazz => JavaType.Class => `java.util.ArrayList`
                    type => JavaType.Parameterized => `java.util.ArrayList<java.lang.String>`
                )
type => JavaType.MethodType => `java.util.ArrayList{name=<constructor>,returnType ...`
)
```

This change is an attempt at setting the appropriate types.

src/test/java/org/openrewrite/kotlin/AssertionsTest.java
```diff
@@ -58,4 +59,9 @@ class A {
           )
         );
     }
+
+    @Test
+    void disablePublishingSnapshots() {
```
traceyyoshima, line 64: @knutwannheden please feel free to remove this. I don't know if there are any changes that would affect the SAAS, so I added this as a safeguard.

src/test/java/org/openrewrite/kotlin/tree/ArrayTest.java
```diff
@@ -94,7 +94,7 @@ void conditionalArrayAccess() {
         );
     }
 
-    @ExpectedToFail("Fixed by PSI-based-parser")
+    @ExpectedToFail("array[0]++ is mapped to 9 Fake elements. Find a way to retrieve the correct FIR.")
```
traceyyoshima, line 97: I don't know how often this will happen. Combinations of syntax sugar that are de-sugared will result in many fake elements. These types of scenarios would need to be mapped manually. We've added an exception in the `PsiElementAssocations` to reveal when we don't find the correct element.

traceyyoshima (commented):
@knutwannheden for review

PR #361, opened 2023-11-06
Fix null JavaClassifierType case

kunli2:
Got this parsing error when parsing repo `dgs-codegen` locally.
Caused by null `JavaClassifierType.classifier`.
Please take the test case in this PR as an example.


traceyyoshima (approved):
Lgtm

PR #364, opened 2023-11-07
New assertion to verify that parser parses whitespace in all locations

knutwannheden:
Adds a new assertion that can be used as an after recipe for any test case. It will insert a comment into every space location, serialize the source, parse it again, and finally serialize it again. Then it compares the two serialized sources to check that they are identical.

Note that a comment in a specific space location is not necessarily parsed back into the same space location again, as some space locations are overlapping / "unreachable" by the parser. But the serialized result should still be identical.


src/main/java/org/openrewrite/kotlin/Assertions.java
```diff
@@ -146,27 +157,89 @@ public static SourceSpecs srcTestKotlin(SourceSpecs... kotlinSources) {
     }
 
     private static void acceptSpec(Consumer<SourceSpec<K.CompilationUnit>> spec, SourceSpec<K.CompilationUnit> kotlin) {
-        Consumer<K.CompilationUnit> userSuppliedAfterRecipe = kotlin.getAfterRecipe();
-        kotlin.afterRecipe(userSuppliedAfterRecipe::accept);
-        isFullyParsed().andThen(spec).accept(kotlin);
+//        Consumer<K.CompilationUnit> consumer = kotlin.getAfterRecipe().andThen(isFullyParsed()).andThen(spaceConscious(kotlin));
```
knutwannheden, line 160: By commenting in this line and commenting out the next line, the assertion can for local testing be enabled globally.

src/test/java/org/openrewrite/kotlin/tree/AnonymousFunctionTest.java
```diff
@@ -23,6 +23,13 @@
 
 class AnonymousFunctionTest implements RewriteTest {
 
+//    @Override
+//    public void defaults(RecipeSpec spec) {
+//        spec
+//          .recipe(Recipe.noop())
+//          .allSources(spaceConscious());
+//    }
+
```
knutwannheden, line 32: This is how the assertion can be enabled for all test cases in a test class. But if individual test cases register their own after recipes, then this one won't be applied and would have to be chained manually.

src/test/java/org/openrewrite/kotlin/tree/AnonymousFunctionTest.java
```diff
@@ -31,6 +38,7 @@ void noArgs() {
             """
               val alwaysTrue =  fun   (    ) =  true
               """
+//            , spec -> spec.afterRecipe(spaceConscious(spec))
```
knutwannheden, line 41: This is how the assertion would be enabled for a single test case.

kunli2 (approved):

Issue #365, opened 2023-11-08
Handle trailing comment of KtFunctionType


kunli2:
Didn't find a place to store the trailing comment of `KtFunctionType` in `K.FunctionType`, we probably want to add a suffix field, ot change `TypedTree returnType` to `JRightPadded<TypedTree>`

Example test 
```java
    @Test
    @Issue("https://github.com/openrewrite/rewrite-kotlin/issues/275")
    void parenthesizedNullableType() {
        rewriteRun(
          kotlin(
            """
              val v: ((Any) -> Any /*c0*/  ) ? = null
              """
          )
        );
    }
```

PR #372, opened 2023-11-10
Fix trailing comment of K.FunctionType

kunli2:
fixes https://github.com/openrewrite/rewrite-kotlin/issues/365

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/tree/K.java
+++ b/src/main/java/org/openrewrite/kotlin/tree/K.java
@@ -1023,7 +1023,7 @@ public interface K extends J {
 
         public FunctionType(UUID id, Space prefix, Markers markers, List<Annotation> leadingAnnotations,
                             List<Modifier> modifiers, @Nullable JRightPadded<NameTree> receiver,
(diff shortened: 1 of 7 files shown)


src/main/java/org/openrewrite/kotlin/tree/K.java
```diff
@@ -1036,9 +1036,8 @@ public FunctionType(UUID id, Space prefix, Markers markers, List<Annotation> lea
         }
 
         public Space getPrefix() {
-            // For backwards compatibility with older LST before there was a prefix field
             //noinspection ConstantConditions
-            return prefix == null ? returnType.getPrefix() : prefix;
+            return prefix == null ? returnType.getElement().getPrefix() : prefix;
```
knutwannheden, line 1040: I suspect we actually want a physical `prefix` field. But I am also not 100% sure, so I suggest we ask @sambsnyd for guidance.

knutwannheden (commented):
@kunli2 Apart from the thing I am unsure about, I think this looks good.

PR #383, opened 2023-11-11
Catch exceptions and create a J.Unknown in visitKtFile.

traceyyoshima:
Changes:
- Exceptions thrown while creating statements are caught and converted to a J.Unknown with a ParseExceptionResult marker.

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
+++ b/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
@@ -39,6 +39,7 @@ import org.jetbrains.kotlin.psi.psiUtil.PsiUtilsKt;
 import org.jetbrains.kotlin.types.Variance;
 import org.openrewrite.ExecutionContext;
 import org.openrewrite.FileAttributes;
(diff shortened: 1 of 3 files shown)


src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
```diff
@@ -1644,7 +1647,22 @@ public J visitKtFile(KtFile file, ExecutionContext data) {
...
         List<KtDeclaration> declarations = file.getDeclarations();
         for (KtDeclaration declaration : declarations) {
-            Statement statement = convertToStatement(declaration.accept(this, data));
+            Statement statement;
+            try {
+                statement = convertToStatement(declaration.accept(this, data));
+            } catch (Exception e) {
+                statement = new J.Unknown(
+                        randomId(),
+                        deepPrefix(declaration),
```
traceyyoshima, line 1656: The prefix of the declaration is retrieved through the prefix method.
Then a substring of the file text is created using PsiUtilsKt.getStartOffsetSkippingComments(declaration) and the declarations endoffset.

Issue #404, opened 2023-11-15
Parsing error: Idempotent print issue on destructuring variable declaration with types


traceyyoshima:
```java
    @Test
    void destructuringWithTypes() {
        rewriteRun(
          kotlin(
            """
              fun example ( ) {
                val ( a : Int , b : String? ) = Pair ( 1 , "Two" )
              }
              """
          )
        );
    }
```

kunli2:
`K.DestructuringDeclaration` does not have a space for variables type.
we may want to update `K.DestructuringDeclaration` model.
One option I can see is to change `K.DestructuringDeclaration#assignments;` from `JContainer<J.VariableDeclarations.NamedVariable>` type to `JContainer<Statement>` and each statement will be a `J.VariableDeclaration`.

@traceyyoshima @knutwannheden for insights.

traceyyoshima:
> `K.DestructuringDeclaration` does not have a space for variables type. we may want to update `K.DestructuringDeclaration` model. One option I can see is to change `K.DestructuringDeclaration#assignments;` from `JContainer<J.VariableDeclarations.NamedVariable>` type to `JContainer<Statement>` and each statement will be a `J.VariableDeclaration`.
> 
> @traceyyoshima @knutwannheden for insights.

I think we should deprecate `JContainer<J.VariableDeclarations.NamedVariable> assignments;`, and make it nullable. Add the appropriate backward compatibility, then add `JContainer<Statement> destructAssignments;`. In most cases, we'll create a new `J.VariableDeclarations` like in `J.MethodDeclaration`. The `Statement` will allow us to handle any other cases that we may discover, and we have precedence for Statement containers already.

kunli2:
yeah, agree, using `JContainer<J.VariableDeclaration>` or `JContainer<Statement>` both works for this case, but I concern that we may discover more cases that might require `JContainer<Statement>` in the future.

PR #426, opened 2023-11-16
Update K.DestructuringDeclaration model to support restructuring declaration with types

kunli2:
Fixes https://github.com/openrewrite/rewrite-kotlin/issues/404
As discussed in the  https://github.com/openrewrite/rewrite-kotlin/issues/404, we decided to update the `K.DestructuringDeclaration` model by adding `JContainer<Statement>` to support restructuring declaration with types

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
+++ b/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
@@ -2302,29 +2302,6 @@ public class KotlinTreeParserVisitor extends KtVisitor<J, ExecutionContext> {
         );
     }
 
-    private J.MethodInvocation buildSyntheticDestructInitializer(int id) {
-        J.Identifier name = new J.Identifier(
-                randomId(),
-                Space.SINGLE_SPACE,
-                Markers.EMPTY,
-                emptyList(),
-                "component" + id,
-                null,
-                null
-        );
-
-        return new J.MethodInvocation(
-                randomId(),
-                Space.EMPTY,
-                Markers.EMPTY,
-                null,
-                null,
-                name,
-                JContainer.empty(),
-                null
-        );
-    }
-
     @Override
     public J visitDotQualifiedExpression(KtDotQualifiedExpression expression, ExecutionContext data) {
         assert expression.getSelectorExpression() != null;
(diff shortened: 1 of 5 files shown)


src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
```diff
@@ -1669,7 +1668,7 @@ public J visitKtFile(KtFile file, ExecutionContext data) {
                         Markers.EMPTY,
                         new J.Unknown.Source(
                                 randomId(),
-                                EMPTY,
+                                Space.EMPTY,
```
traceyyoshima, line 1671: I'm just curious about this change since other parsers typically use `import static org.openrewrite.java.tree.Space.EMPTY;`. Has that changed, or was this an aesthetic preference?

src/main/java/org/openrewrite/kotlin/tree/K.java
```diff
@@ -911,12 +927,22 @@ public K.DestructuringDeclaration.Padding getPadding() {
         public static class Padding {
             private final K.DestructuringDeclaration t;
 
+            @Deprecated
```
traceyyoshima, line 930:
```suggestion
            @Deprecated
            @Nullable
```

src/main/kotlin/org/openrewrite/kotlin/internal/KotlinParserVisitor.kt
```diff
@@ -1393,7 +1393,8 @@ class KotlinParserVisitor(
                 prefix,
                 Markers.EMPTY,
                 variableDeclarations,
-                JContainer.build(before, vars, Markers.EMPTY)
+                // JContainer.build(before, vars, Markers.EMPTY)
+                JContainer.build(before, emptyList(), Markers.EMPTY) // pass compile
```
traceyyoshima, line 1397: Ah, I didn't think this still existed -- could you remove the `KotlinParserVisitor` in a separate PR?

traceyyoshima (changes requested):
You may add `"DeprecatedIsStillUsed"` to the suppression on the class.

Please, update the `KotlinVisitor` as well.

src/main/kotlin/org/openrewrite/kotlin/internal/KotlinParserVisitor.kt
kunli2: Sure, why not.

src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
kunli2: yeah, just aesthetic preference, since I think it will make it more clear to differentiate with other `EMPTY`.

kunli2:
> KotlinVisitor

since there are some other warnings that are not related to this PR,   I will have a separate PR to clean up all remaining warnings

PR #440, opened 2023-11-20
Change type/package secondary constructor issues. Change type annotation issue

rogerpeckham-toast:
## What's changed?

This is a PR for #433. Currently it only has unit tests to reproduce the issue. It also has two more unit tests, one for basically the same issue but with change package, the other for an issue with change type on an annotation.
### Checklist
- [ ] I've added unit tests to cover both positive and negative cases
- [ ] I've added the license header to any new files through `./gradlew licenseFormat`
- [ ] I've used the IntelliJ IDEA auto-formatter on affected files

### DIFF
--- a/src/test/java/org/openrewrite/kotlin/ChangePackageTest.java
+++ b/src/test/java/org/openrewrite/kotlin/ChangePackageTest.java
@@ -166,4 +166,36 @@ class ChangePackageTest implements RewriteTest {
           )
         );
     }
+
+    @Test
+    void changePackageSecondaryConstructor() {
+        rewriteRun(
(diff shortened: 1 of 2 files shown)


kunli2:
Thanks @rogerpeckham-toast for creating those tests, that will be helpful to us in fixing the issues.
we will let you know once it's fixed.

traceyyoshima:
@rogerpeckham-toast thank you for the PR! I'm going to merge the tests, so you have credit in the git history!

traceyyoshima (approved):
LGTM, thank you for reproducing the issue!

Issue #442, opened 2023-11-20
Parsing error: UOE on KotlinTreeParserVisitor#visitIntersectionType


traceyyoshima:
```java
              import java.util.*
              
              @Suppress("UNUSED_PARAMETER")
              fun <T : Any?> test(n: Optional<T & Any> = Optional.empty<T>()) {}
```

```
java.lang.UnsupportedOperationException: TODO
  org.openrewrite.kotlin.internal.KotlinTreeParserVisitor.visitIntersectionType(KotlinTreeParserVisitor.java:712)
  org.openrewrite.kotlin.internal.KotlinTreeParserVisitor.visitIntersectionType(KotlinTreeParserVisitor.java:71)
  org.jetbrains.kotlin.psi.KtIntersectionType.accept(KtIntersectionType.kt:22)
  org.openrewrite.kotlin.internal.KotlinTreeParserVisitor.visitTypeReference(KotlinTreeParserVisitor.java:3062)
  org.openrewrite.kotlin.internal.KotlinTreeParserVisitor.visitTypeReference(KotlinTreeParserVisitor.java:71)
  org.jetbrains.kotlin.psi.KtTypeReference.accept(KtTypeReference.kt:38)
  org.openrewrite.kotlin.internal.KotlinTreeParserVisitor.visitTypeProjection(KotlinTreeParserVisitor.java:1411)
  org.openrewrite.kotlin.internal.KotlinTreeParserVisitor.visitTypeProjection(KotlinTreeParserVisitor.java:71)
  ...
```

PR #454, opened 2023-11-21
Support intersection type

kunli2:
fixes https://github.com/openrewrite/rewrite-kotlin/issues/442
To support the intersection type (like `T   &    Any` below),  it can be parsed to a `J.TypeParameter`, and this requires `rewrite` model update https://github.com/openrewrite/rewrite/pull/3718.

```kotlin
              import java.util.*
              @Suppress("UNUSED_PARAMETER")
              fun <T : Any?> test(n: Optional <  T   &    Any > = Optional.empty<T>()) {}
```

Alternatively, we can parse `T   &    Any` to a `K.Binary` instead of a `J.TypeParameter` with a new binary type.


kunli2:
slack discussion: https://moderneinc.slack.com/archives/C04CS5N0PJN/p1700529762668109
we will use `J.IntersectionType` to support Kotlin intersection type, so closing out this PR.

Issue #453, opened 2023-11-21
Parsing error: Unsupported call type in PsiElementAssociations#getCallType


traceyyoshima:
I attempted to simplify the code more, but the `FirArrayOfCall` would only appear for this:
```kotlin
annotation class Ann(
    val test: Test
)
annotation class Test ( val arg: Array<String> )
@Ann(test = Test(
    arg = arrayOf("")
))
fun use() {
}
```

PR #458, opened 2023-11-21
Fix exception of `arrayOf` call

kunli2:
fixes https://github.com/openrewrite/rewrite-kotlin/issues/453

### DIFF
--- a/src/main/kotlin/org/openrewrite/kotlin/internal/PsiElementAssociations.kt
+++ b/src/main/kotlin/org/openrewrite/kotlin/internal/PsiElementAssociations.kt
@@ -123,7 +123,8 @@ class PsiElementAssociations(val typeMapping: KotlinTypeMapping, val file: FirFi
                     }
 
                     is FirArrayOfCall -> {
-                        // todo, Didn't find a type from FirArrayOfCall
+                        // `FirArrayOfCall` is not a `FirFunctionCall`, so a `JavaType$Method` is impossible.
+                        // The expression contains a type ref of the parameterized type, but cannot be added to a `MethodInvocation`.
                         null
                     }
                     else -> {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/openrewrite/kotlin/internal/PsiElementAssociations.kt
```diff
@@ -197,6 +201,7 @@ class PsiElementAssociations(val typeMapping: KotlinTypeMapping, val file: FirFi
         val fir = primary(psi) ?: return null
         return when (fir) {
             is FirResolvedQualifier -> ExpressionType.QUALIFIER
+            is FirArrayOfCall -> ExpressionType.METHOD_INVOCATION
```
traceyyoshima, line 204: I'll look more into this tomorrow.

`FirArrayOfCall` implements `FirCall` and is not a `FirFunctionCall`. So, from the `FIR` it isn't a `METHOD_INVOCATION`.

There is however, an `arrayOf` function in `Library.kt`. But `Library.kt` isn't resolved until the IR phase, so we may not be able to get the functions type.

`FirArrayOfCall` does contain a `typeRef` since it is an expression and will return the parameterized type from the functions cast: `public inline fun <T> arrayOf(vararg elements: T): Array<T> = elements as Array<T>`. So, using type(firArrayOfCall) will work, but a method invocation won't.

src/main/kotlin/org/openrewrite/kotlin/internal/PsiElementAssociations.kt
```diff
@@ -122,6 +122,10 @@ class PsiElementAssociations(val typeMapping: KotlinTypeMapping, val file: FirFi
                         }
                     }
 
+                    is FirArrayOfCall -> {
+                        // todo, Didn't find a type from FirArrayOfCall
```
traceyyoshima, line 126:
```suggestion
                        // `FirArrayOfCall` is not a `FirFunctionCall`, so a `JavaType$Method` is impossible.
                        // The expression contains a type ref of the parameterized type, but cannot be added to a `MethodInvocation`.
```

traceyyoshima (approved):
LGTM!

`FirArrayCall` cannot be converted to a method type. Other common uses of `arrayOf(..)` are converted to a `FirFunctionCall`.

Issue #464, opened 2023-11-23
PsiElementAssociations#primary() can return the wrong associations


traceyyoshima:
Issue:
   The PSI-based parser calls `type()` through `visitSimpleNameExpression` on names that cannot have types and/or names not associated with a FIR element. `PsiElementAssociations#fir` climbs the PSI#parent nodes until a FIR association is detected.

On parameterized type names:
    `List` from `List<String>`. The identifier List will have a type of `...List<kotlin.String>` instead of `...List`.

PR #471, opened 2023-11-28
Fixed various types on J.Identifier. Use mapType to fix types based on the tree.

traceyyoshima:
fixes https://github.com/openrewrite/rewrite-kotlin/issues/464

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
+++ b/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
@@ -947,7 +947,7 @@ public class KotlinTreeParserVisitor extends KtVisitor<J, ExecutionContext> {
             throw new UnsupportedOperationException("TODO");
         }
 
-        return new J.MethodDeclaration(
+        return mapType(new J.MethodDeclaration(
                 randomId(),
                 deepPrefix(constructor),
                 Markers.build(singletonList(new PrimaryConstructor(randomId()))),
@@ -964,7 +964,7 @@ public class KotlinTreeParserVisitor extends KtVisitor<J, ExecutionContext> {
                 null,
                 null,
                 type
-        );
+        ));
     }
 
     @Override
@@ -1019,7 +1019,7 @@ public class KotlinTreeParserVisitor extends KtVisitor<J, ExecutionContext> {
                     .withMarkers(Markers.EMPTY.addIfAbsent(new OmitParentheses(randomId())));
         }
 
-        return new J.MethodDeclaration(
+        return mapType(new J.MethodDeclaration(
                 randomId(),
                 deepPrefix(accessor),
                 markers,
@@ -1036,7 +1036,7 @@ public class KotlinTreeParserVisitor extends KtVisitor<J, ExecutionContext> {
                 body,
                 null,
                 type
-        );
+        ));
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


traceyyoshima:
Hi @knutwannheden -- this is a draft to fix several type issues in the LST.

I came across an odd issue that I am still trying to figure out how to resolve. Please take a look in case you quickly see a simple solution.

Info:
I discovered that the FIR associated with names of type parameters is the parent element rather than the parameter.
So, for `List<String>`, the name `String` should have a type `kotlin.String`, but is set to `...List<kotlin.String>`.

When there are multiple levels of parameterization, the types become pretty strange. You can see the affect with this test:
```java
        @Test
        void multiDepthParameterizedType() {
            rewriteRun(
              kotlin(
                """
                  val map: Map<Map<Int, Int>, Map<String, String>> = mapOf()
                  """, spec -> spec.afterRecipe(cu -> {
                        @Override
                        public J.Identifier visitIdentifier(J.Identifier identifier, Integer integer) {
                            System.out.println(identifier.getType().toString());
                            return super.visitIdentifier(identifier, integer);
                        }
                    }.visit(cu, 0);
                })
              )
            );
        }
```

traceyyoshima:
Additional info:
The idea is to use the J tree and types from the Java compiler to set the appropriate types in the K tree.
Then, by detecting unexpected mappings due to PSI to FIR associations, we can add markers to the tree and use the SAAS to identify the problems and proactively fix type attribution problems.

knutwannheden:
@traceyyoshima I would have expected that we in the element associations can have a special case for type parameters and then navigate doen to the corresponding type parameter element in the FIR. Does that not work?

traceyyoshima:
> @traceyyoshima I would have expected that we in the element associations can have a special case for type parameters and then navigate doen to the corresponding type parameter element in the FIR. Does that not work?

IIRC, the `type` method is called in `visitSimpleName`, which is some form of name reference, which doesn't have context to a type parameter. But yes, possibly, each of the methods that creates a`J.ParameterizedType` can use the list argument list and associate the appropriate class type. I'll need to think through different compositions of parameterized types. I don't know enough about the PSI to be certain.

src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
```diff
@@ -3357,25 +3338,146 @@ private J.If.Else buildIfElsePart(KtIfExpression expression) {
     /*====================================================================
      * Type related methods
      * ====================================================================*/
-    private <T extends J> T mapType(T tree) {
-        return mapType(tree, null);
-    }
-
     @SuppressWarnings("unchecked")
-    private <T extends J> T mapType(T tree, @Nullable JavaType type) {
+    private <T extends J> T mapType(T tree) {
```
knutwannheden, line 3342: @traceyyoshima Can you briefly explain why we need / want to do it this way rather than trying to set the appropriate type already at construction time? I think we should then also add a comment here to that extent, because to me this isn't entirely clear.

knutwannheden (approved):

src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
knutwannheden: Since this method is really just one switch and we often at the call aite already know the object's type, I would have overloaded the method for every type. I assume that we won't always know the type at compile time at the call site and will thus still require this method to delegate to the others.

I think we should however also think about a solution where we can vonstruct the correct type attribution to begin with. Possibly by doing something similar as we did for the generic type arguments in the element associations. Alternatively, if we don't think this is feasible, we should just add a brief comment here explaining why this fixup is required here.

traceyyoshima: Ah, that does sound clearer -- I'll add overloads and comments about the changes.

AFAIK, there is not an alternative yet, mostly because we don't know what the mis-mappings are. The changes are meant to allow us to detect type mapping issues so that we can fix them proactively.

PR #478, opened 2023-11-29
Added detection of annotation constructors to create J.Annotation in visitCallExpression.

### LINKED ISSUE #475: visitAnnotation method is not called for nested annotation
## What version of OpenRewrite are you using?
I am using

- OpenRewrite v8.9.9
- OpenRewrite Kotlin v1.7.7

## What is the smallest, simplest way to reproduce the problem?

```java
import org.junit.jupiter.api.Test;
import org.openrewrite.java.ChangeAnnotationAttributeName;
import org.openrewrite.java.JavaParser;
import org.openrewrite.kotlin.KotlinParser;
import org.openrewrite.test.RewriteTest;

import static org.openrewrite.java.Assertions.java;
import static org.openrewrite.kotlin.Assertions.kotlin;

public class ChangeAnnotationAttributeNameTest implements RewriteTest {

    @Test
    public void run() {
        rewriteRun(
                spec -> spec.recipe(new ChangeAnnotationAttributeName(
                                "org.junit.jupiter.api.Tag",
                                "value",
                                "newValue"
                        ))
                        .parser(KotlinParser.builder().classpath("junit-jupiter-api")),
                kotlin(
                        """
                                package sample
                                 
                                 import org.junit.jupiter.api.Tag
                                 import org.junit.jupiter.api.Tags
                                 
                                 class SampleTest {
                                 
                                     @Tags(
                                         value = [
                                             Tag(value = "Sample01"),
                                             Tag(value = "Sample02"),
                                         ]
                                     )
                                     fun run() {
                                     }
                                 }
                                """,
                        """
                                package sample
                                 
                                 import org.junit.jupiter.api.Tag
                                 import org.junit.jupiter.api.Tags
                                 
                                 class SampleTest {
                                 
                                     @Tags(
                                         value = [
                                             Tag(newValue = "Sample01"),
                                             Tag(newValue = "Sample02"),
                                         ]
                                     )
                                     fun run() {
                                     }
                                 }
                                """
                )
        );
    }
}
```

traceyyoshima:
Changes:

- `ChangeAnnotationAttributeName` will properly change attributes of J.Annotations.
- J.Annotation is created when used as a value (when type attribution is available) in visitCallExpression.
- AnnotationConstructor marker and deprecated `Modifier` marker. Replaced use of `Modifier` with `AnnotationConstructor`.

fixes #475

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
+++ b/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
@@ -1689,7 +1689,7 @@ public class KotlinTreeParserVisitor extends KtVisitor<J, ExecutionContext> {
 
         for (KtAnnotationEntry ktAnnotationEntry : annotationEntries) {
             J.Annotation anno = (J.Annotation) ktAnnotationEntry.accept(this, data);
-            anno = anno.withMarkers(anno.getMarkers().addIfAbsent(new Modifier(randomId())));
+            anno = anno.withMarkers(anno.getMarkers().addIfAbsent(new AnnotationConstructor(randomId())));
             rpAnnotations.add(padRight(anno, Space.EMPTY));
         }
 
@@ -1727,7 +1727,7 @@ public class KotlinTreeParserVisitor extends KtVisitor<J, ExecutionContext> {
             J.Annotation argAnno = new J.Annotation(
                     randomId(),
                     Space.EMPTY,
-                    Markers.EMPTY.addIfAbsent(new Modifier(randomId())),
+                    Markers.EMPTY.addIfAbsent(new AnnotationConstructor(randomId())),
                     (NameTree) requireNonNull(annotationEntry.getCalleeExpression()).accept(this, data),
                     annotationEntry.getValueArgumentList() != null ? mapValueArguments(annotationEntry.getValueArgumentList(), data) : null
             );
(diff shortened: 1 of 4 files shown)


src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
```diff
@@ -2339,43 +2361,50 @@ public J visitDestructuringDeclaration(KtDestructuringDeclaration multiDeclarati
     @Override
     public J visitDotQualifiedExpression(KtDotQualifiedExpression expression, ExecutionContext data) {
         assert expression.getSelectorExpression() != null;
+        Space prefix = deepPrefix(expression);
```
traceyyoshima, line 2364: @kunli2 There were a few missing calls to deepPrefix, which I added, and it seems fine.
But I don't know when different prefix methods need to be used.

Is it okay to call deepPrefix on each returned element?

knutwannheden (approved):
I think this looks good! It would however be good if we could add a test that also verifies, that when parsing a nested annotation (without `@`) that the produced LST also looks correct and is properly type attributed.

traceyyoshima:
> I think this looks good! It would however be good if we could add a test that also verifies, that when parsing a nested annotation (without `@`) that the produced LST also looks correct and is properly type attributed.

Sounds good! Luckily, the engineer who reported the issue created a [PR](https://github.com/openrewrite/rewrite-kotlin/commit/63bf317d696da6c2f33b087136b88da3a730065c) which has already been merged. The test confirms the source is parsed and printed idempotent, and the `ChangeAnnotationAttributeName` implies the J.Annotations have the correct types, since the attributed are updated.

PR #487, opened 2023-11-30
Map types in `PsiElementAssociations`

### LINKED ISSUE #473: Incorrect type associated from the PSI to the FIR for parameterized types.
TLDR: The type parameters will have the wrong JavaType.

Info:
When the PSI visits the name of the type parameters, an FIR association will not be found.
The PSI will traverse the parent nodes until it matches the most outer parameterized type.
I.E. The name `String` in `List<String>` will climb from `String` up to `List<String>` or the name `Int` of `Map<Map<Int, Int>, Map<String, String>>` will climb up to `Map<Map<Int, Int>, Map<String, String>>`.

Note: This may be caused by referencing the wrong PSI element and might be fixable using PSI-Utils from each visit that creates type parameters for parameterized types. The `elementMap` in PsiElementAssociations should be checked first to determine if we only need to revise the mapping instead of anything more complex.

```java
        @Test
        void multiDepthParameterizedType() {
            rewriteRun(
              kotlin(
                """
                  val map: Map<Map<Int, Int>, Map<String, String>> = mapOf()
                  """, spec -> spec.afterRecipe(cu -> {
                        @Override
                        public J.Identifier visitIdentifier(J.Identifier identifier, Integer integer) {
                            // Identifier should be a JavaType.Class, but it will be a JavaType.Parameterized.
                            return super.visitIdentifier(identifier, integer);
                        }
                    }.visit(cu, 0);
                })
              )
            );
        }
```

knutwannheden:
The types themselves are not FIR elements and thus not visited by the visitor used to populate the associations map. However, these can be visited separately, so that they later can be retrieved using `type()`.

fixes #473

### DIFF
--- a/src/main/kotlin/org/openrewrite/kotlin/internal/PsiElementAssociations.kt
+++ b/src/main/kotlin/org/openrewrite/kotlin/internal/PsiElementAssociations.kt
@@ -71,8 +71,14 @@ class PsiElementAssociations(val typeMapping: KotlinTypeMapping, val file: FirFi
                 data: MutableMap<PsiElement, MutableList<FirInfo>>
             ) {
                 super.visitResolvedTypeRef(resolvedTypeRef, data)
-                if (resolvedTypeRef.type is ConeClassLikeType && resolvedTypeRef.type.typeArguments.isNotEmpty() && resolvedTypeRef.psi is KtTypeReference) {
-                    visitType(resolvedTypeRef.type, resolvedTypeRef.psi as KtTypeReference)
+                if (resolvedTypeRef.psi is KtTypeReference) {
+                    if (resolvedTypeRef.type is ConeClassLikeType) {
+                        if (resolvedTypeRef.type.typeArguments.isNotEmpty() && resolvedTypeRef.psi is KtTypeReference) {
+                            visitType(resolvedTypeRef.type, resolvedTypeRef.psi as KtTypeReference)
+                        }
+                    } else if (resolvedTypeRef.type is ConeTypeParameterType) {
+                        visitType(resolvedTypeRef.type, resolvedTypeRef.psi as KtTypeReference)
+                    }
                 }
             }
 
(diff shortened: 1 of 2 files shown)


knutwannheden:
@traceyyoshima This is basically what I meant in [my comment](https://github.com/openrewrite/rewrite-kotlin/pull/471#issuecomment-1829138412). Only here the element associations would "proactively" add an association for the type, rather than at lookup time try to navigate the trees to find it again.

I've created this as a draft, because I am a bit unsure if this is generic enough with the two parent calls in the `type()` method and if we cover all cases like this. Also, we should adjust / add tests for this. But with the test from your PR, the output is now what you were expecting to see. I.e. the type parameters don't have the owning type as their type anymore.

I assigned this PR to you, as I think you are in a better position to take this over the finishing line.

src/main/kotlin/org/openrewrite/kotlin/internal/PsiElementAssociations.kt
```diff
@@ -65,9 +67,65 @@ class PsiElementAssociations(val typeMapping: KotlinTypeMapping, val file: FirFi
...
+
+            private fun visitType(firType: ConeTypeProjection, psiType: KtTypeReference) {
+                if (firType is ConeClassLikeType) {
+                    val psiTypeArguments = psiType.typeElement!!.typeArgumentsAsTypes
+                    if (psiTypeArguments.size != firType.typeArguments.size) {
+                        return
+                    }
+
+                    for ((index, typeArgument) in firType.typeArguments.withIndex()) {
+                        val psiTypeArgument = psiTypeArguments[index] ?: continue
```
traceyyoshima, line 94: Wildcards like `*` in `List<*>` will not resolve to a PSI here, so the identifier for `*` will be missing a type.
I don't think it's an issue, because `J.Wildcard` implies the type regardless of the presence of a `JavaType$GenericTypeVariable`.

knutwannheden: Having this as a comment in the code would be helpful I think.

Issue #477, opened 2023-11-29
Improve quality of LST: replace applicable markers with LST elements


traceyyoshima:
Replace:
- [x] AnnotationUseSite:
        The `J` tree expects `J.Annotation`(s), a final class. In `K`, there is an additional suffix after the use site annotation like `@file : ...`. This change may require an update to `J.Annotation` since various classes require it.
- [x] IsNullable:
        Nullable is a generally applicable concept across `J` and every implementation of `J`, and may be generally fixable. https://github.com/openrewrite/rewrite-kotlin/pull/489
https://github.com/openrewrite/rewrite-kotlin/pull/497
- [x] IsNullSafe:
        Possible solution: Add an IsNullSafe `Expression` pass-through to contain the applicable expression.
- [x] Modifier: The modifier marker was supposed to be replaced in the PSI parser with the `J.Modifier#keyword`, and appears to still be in use.
- [ ] TypeReferencePrefix
        Possible solution: Add a TypeReference `TypeTree` tree to contain the applicable expression.

Update:
- [x] Elvis:
       This marker exists in both Groovy and Kotlin, and will likely be reused elsewhere. We should probably add this to rewrite-java and use the reuse the marker.

kunli2:
`Elvis` marker is not used anymore in Kotlin and has been changed a `K.Binary`, so we can deprecate it directly here.

kunli2:
I checked `IsNullSafe` marker again, there are 3 usage scenarios 
1. `J.FieldAccess`, example: `test   ?. property`, the space before `?.` is handled by leftPadded
2. `J.MethodInvocation`, example: 
```
              val filter = null   ?. let { _ ->
                  { false
                  }
              }
```
   the space before `?.` is handled by suffix of the select (RightPadded)

3. `J.TypeCast`, example `val map = mapOf(Pair("one", 1))   as? Map<*, *>`, the space before  `?.` is handled by the prefix of `J.TypeCast.clazz`

so I think the prefix in this marker is not necessary, we can remove it or deprecate it and keep the marker there.
@knutwannheden @traceyyoshima  what do you think?

knutwannheden:
Let's also try to be careful to not break LST deserialization as we do these refactorings. In moderne-ast-write we can set up regression tests, so that we know if we need to customize deserialization or not.

PR #489, opened 2023-11-30
Deprecate `IsNullable` marker and use `K.Unary` instead

kunli2:
Use `K.Unary` to present `nullableType` instead of `IsNullable` marker

To support this, we need to change `K.Unary` extends `TypeTree` instead of `TypedTree`

part of https://github.com/openrewrite/rewrite-kotlin/issues/477

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
+++ b/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
@@ -800,7 +800,16 @@ public class KotlinTreeParserVisitor extends KtVisitor<J, ExecutionContext> {
 
     @Override
     public J visitNullableType(KtNullableType nullableType, ExecutionContext data) {
-        TypeTree typeTree = (TypeTree) requireNonNull(nullableType.getInnerType()).accept(this, data);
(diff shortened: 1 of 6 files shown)


src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
```diff
@@ -68,7 +68,7 @@
 /**
  * PSI based parser
  */
-@SuppressWarnings("ConstantValue")
+@SuppressWarnings({"ConstantValue", "DataFlowIssue", "SameParameterValue", "unused"})
```
traceyyoshima, line 71: Please remove DataFlowIssue suppression.
For future reference, it causes potential NPEs not to be shown in the IDE, so we shouldn't add it.

traceyyoshima (commented):
I don't think we should merge this, there is a note about a general fix in the issue: `Nullable is a generally applicable concept across J and every implementation of J, and may be generally fixable.`

Nullability applies to Java, and multiple implementations of `J`, so we should consider a general fix first. I think @knutwannheden is already thinking of this.

kunli2:
slack discussion : https://moderneinc.slack.com/archives/C067J83AFHB/p1701369801264339

PR #491, opened 2023-12-01
Deprecate the `prefix` field of `IsNullSafe` marker since it's uncessery

kunli2:
There are total 3 usage scenarios of `IsNullSafe` marker.
1. `J.FieldAccess`, example: `test   ?. property`, the space before `?.` is handled by leftPadded
2. `J.MethodInvocation`, example: 
```
              val filter = null   ?. let { _ ->
                  { false
                  }
              }
```
   the space before `?.` is handled by suffix of the select (RightPadded)

3. `J.TypeCast`, example `val map = mapOf(Pair("one", 1))   as? Map<*, *>`, the space before  `?.` is handled by the prefix of `J.TypeCast.clazz`

so the prefix in this marker is not necessary, we can just deprecate the `prefix` field and keep the marker there.

this PR is related to https://github.com/openrewrite/rewrite-kotlin/issues/477

### DIFF
--- a/src/test/java/org/openrewrite/kotlin/format/TabsAndIndentsTest.java
+++ b/src/test/java/org/openrewrite/kotlin/format/TabsAndIndentsTest.java
@@ -2250,13 +2250,17 @@ class TabsAndIndentsTest implements RewriteTest {
             """
               fun f(): String? {
                   val values = (listOf("") as List<String>?)
-                              ?.map { it }
+                      ?.map { it }
                           ?: return null
                   return values.joinToString("")
               }
               """
           )
         );
+    }
+
+    @Test
+    void ternaryIndentation2() {
         rewriteRun(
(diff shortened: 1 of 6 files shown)


src/test/java/org/openrewrite/kotlin/format/TabsAndIndentsTest.java
```diff
@@ -2250,28 +2250,32 @@ fun f(): String? {
             """
               fun f(): String? {
                   val values = (listOf("") as List<String>?)
-                              ?.map { it }
+                      ?.map { it }
```
traceyyoshima, line 2253: Why do tests like this change if the prefix is unused and unnecessary?
There there code else where like in formatting that references the markers?

traceyyoshima: Ahh, okay, that makes sense. The prefix is added in both positions and printed twice. Thank you!

traceyyoshima (approved):

PR #499, opened 2023-12-04
Reduce STDERR in AutoDetectTest

timtebeek:
I suspect these tests carried over from Java, but hadn't fully been Kotlin-ized yet.


traceyyoshima:
> Any idea why those methods might lack the static modifier after going through the Kotlin parser @knutwannheden ?
Taking a look

Issue #493, opened 2023-12-01
ChangeType fails on imports with escaped names


traceyyoshima:
Imports are created through TypeTree.build, which does not support escaped identifiers.

- [x] Add support for escaped identifiers to TypeTree.build.
- [x] Update type on the import with type and detect `JavaType$ParameterizedType` to set the `JavaType$Class`.

traceyyoshima:
```java
    @Test
    void changeImport() {
        rewriteRun(
          kotlin(
            """
              package a.b
              class Original
              """),
          kotlin(
            """
              import a.b.`Original`
              
              class A {
                  val type : Original = Original()
              }
              """,
            """
              import x.y.`Target`
              
              class A {
                  val type : Target = Target()
              }
              """
          )
        );
    }
```

PR #501, opened 2023-12-04
Added support for imports with escaped names.

traceyyoshima:
Changes:

- Imports are created with escaped FQNs using Quote markers in `KotlinTreeParserVisitor`.
- Imports will have the resolved types rather than a shallow class.
- AddImport supports escaped FQNs.
- ChangeType will correctly update types with escaped names.

fixes https://github.com/openrewrite/rewrite-kotlin/issues/493


kunli2 (approved):
LGTM

PR #504, opened 2023-12-05
Fixed type attribution on annotation on ConeKotlinType(s).

### LINKED ISSUE #488: Suppress annotations can be assigned the wrong type
TODO:
- Determine if there are any cases where a Suppress annotation has an associated FIR.
    - Retention or a compiler optimization may cause it.
    - If the issue is caused by Retention, we need to fix the type mapping issue for all applicable annotations.
- Prevent incorrect types from being assigned to the annotation.

The misattribution depends on the position of the annotation, in some cases, when the PSI climbs parent trees the parent will not resolve to a type.

```kotlin
val releaseDates: List< /*C0*/ @Suppress  /*C1*/ String> = emptyList()
```

traceyyoshima:
Change:

- Annotations on `ConeKotlinType ` will contain the correct type attribution.

fixes #488


kunli2 (approved):
lgtm

PR #505, opened 2023-12-05
Get rid of `TypeReferencePrefix` marker from `J.VariableDeclarations`

kunli2:
This part of https://github.com/openrewrite/rewrite-kotlin/issues/477
To get rid of  `TypeReferencePrefix` marker from `J.VariableDeclarations`.

Here are 3 `TypeReferencePrefix` use cases
### 1. Variable declaration statement in a block

```kotlin
class Test {
    var t /*C0*/ : /*C1*/ Int = 1 /*C2*/  ; /*C3*/
    fun method() {}
}
```

For this, this PR changes the spaces locations from

```
// "C0" -> Marker of `TypeReferencePrefix`
// "C1" -> prefix of typeExpression
// "C2" -> right padding of variable
```
to
```
// "C0" -> right padding of variable
// "C2" -> right padding of variable of the statement
```

### 2. Variable declaration as Parameters
```kotlin
fun method ( input /*C0*/ : Any = 1 /*C2*/ , x : Int ) {
    return "42"
}
 
```

Change from 
```
// "C0" : marker of `TypeReferencePrefix`
// "C2" : right padding of the statement
```
to
```
// "C0" : right padding of variable
// "C2" : right padding of the statement
```

### 3. Method return type
```kotlin
fun method () /*C3*/ :  Int {
    return 42
}
```
This is not covered by this PR.

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/internal/KotlinPrinter.java
+++ b/src/main/java/org/openrewrite/kotlin/internal/KotlinPrinter.java
@@ -368,10 +368,14 @@ public class KotlinPrinter<P> extends KotlinVisitor<PrintOutputCapture<P>> {
             }
         }
 
-        if (!vd.getVariables().isEmpty()) {
-            J.VariableDeclarations.NamedVariable nv = vd.getVariables().get(0);
+        List<JRightPadded<J.VariableDeclarations.NamedVariable>> rpvs =  vd.getPadding().getVariables();
+        if (!rpvs.isEmpty()) {
+            JRightPadded<J.VariableDeclarations.NamedVariable> rpv = vd.getPadding().getVariables().get(0);
+            J.VariableDeclarations.NamedVariable nv = rpv.getElement();
             beforeSyntax(nv, Space.Location.VARIABLE_PREFIX, p);
             visit(nv.getName(), p);
+            visitSpace(rpv.getAfter(), Space.Location.TYPE_PARAMETERS, p);
+
             if (vd.getTypeExpression() != null) {
                 vd.getMarkers().findFirst(TypeReferencePrefix.class).ifPresent(tref -> visitSpace(tref.getPrefix(), KSpace.Location.TYPE_REFERENCE_PREFIX, p));
                 p.append(":");
@@ -1151,10 +1155,15 @@ public class KotlinPrinter<P> extends KotlinVisitor<PrintOutputCapture<P>> {
             beforeSyntax(multiVariable, Space.Location.VARIABLE_DECLARATIONS_PREFIX, p);
 
             visit(multiVariable.getLeadingAnnotations(), p);
+            boolean isTypeAlias = false;
             for (J.Modifier m : multiVariable.getModifiers()) {
                 visitModifier(m, p);
                 if (m.getType() == J.Modifier.Type.Final) {
                     p.append("val");
+                }
+
+                if (m.getType() == J.Modifier.Type.LanguageExtension && m.getKeyword().equals("typealias")) {
+                    isTypeAlias = true;
                 }
             }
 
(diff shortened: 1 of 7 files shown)


traceyyoshima (commented):
I think it may be simpler to solve `TypeTree` for the `returnType` of `J.MethodDeclaration` and apply that change to each location of a `TypeReference`, since each will be a `TypeTree`.

That will allow auto-formatting to visit a single location and apply the changes uniformly across the entire LST for the space before colons, as well as make it easier for users to make updates with recipes. Do you not agree?

src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
```diff
@@ -4139,6 +4137,26 @@ private static PsiElement findLastChild(@Nullable PsiElement parent, Predicate<P
         return null;
     }
 
+    // Search for the last child that satisfies a given condition. If at any point, a child meets a break condition before the condition is satisfied, the function should return null.
+    @Nullable
+    private static PsiElement findLastChild(@Nullable PsiElement parent, Predicate<PsiElement> condition,  Predicate<PsiElement> breakCondition) {
```
traceyyoshima, line 4142: I suspect existing PSI utils can do this already -- were the existing utilities insufficient?

kunli2: good one, thanks, I will update

traceyyoshima: Sounds good -- as we pay tech debt, I think it'll be helpful to replace most unnecessary methods with PSI utils.

Moving forward, it will be helpful/beneficial to consider what utilities already exist rather than adding new code. Generally, I consider what is available before adding code like findLastChild. It's helpful to do so, since it prevents unnecessary complexity, and uses code that is already tested. I'm fairly certain that findLastChild can be replaced, thanks in advance!

traceyyoshima: > I didn't find one from `PsiTreeUtil` to do the equivalent thing below, @traceyyoshima , did you see any method to support this?

Ah, I can show you how to find utilities again if that helps -- I don't mean to look at `PsiTreeUtil`. As you know, there are many utilities in the Kotlin compiler, including for the PSI. Several forms of utils, utility classes, and classes with extension methods exist.

traceyyoshima:
> Here are 3 `TypeReferencePrefix` use cases

I think there's been a misunderstanding, the reason I mentioned return types on method declarations and that we'll need someone to check the LST for other cases is that there are other positions for TypeTree where a `:` may exist. There is at least one additional case, and I haven't checked on any others.

src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
```diff
@@ -885,7 +885,7 @@ public J visitParameter(KtParameter parameter, ExecutionContext data) {
                 vt
         );
 
-        vars.add(padRight(namedVariable, Space.EMPTY));
+        vars.add(padRight(namedVariable, prefix(parameter.getColon())));
```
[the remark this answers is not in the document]
kunli2: good catch, yes, it can be safely removed, I just verified that and removed.

src/main/java/org/openrewrite/kotlin/internal/KotlinPrinter.java
```diff
@@ -1151,11 +1155,16 @@ public J visitVariableDeclarations(J.VariableDeclarations multiVariable, PrintOu
...
             visit(multiVariable.getLeadingAnnotations(), p);
+            boolean isTypeAlias = false;
             for (J.Modifier m : multiVariable.getModifiers()) {
                 visitModifier(m, p);
                 if (m.getType() == J.Modifier.Type.Final) {
                     p.append("val");
                 }
+
+                if (m.getType() == J.Modifier.Type.LanguageExtension && m.getKeyword().equals("typealias")) {
+                    isTypeAlias = true;
```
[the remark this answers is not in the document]
kunli2: Yes, having a `K.TypeAlias` is something we can do in a separate PR.

knutwannheden (approved):

PR #522, opened 2023-12-11
Update `K.Constructor` model

kunli2:
Use `JLeftPadded<ConstructorInvocation>` rather than separate the Space colon on Constructor.

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/tree/K.java
+++ b/src/main/java/org/openrewrite/kotlin/tree/K.java
@@ -15,6 +15,8 @@
  */
 package org.openrewrite.kotlin.tree;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import lombok.*;
 import lombok.experimental.FieldDefaults;
 import lombok.experimental.NonFinal;
@@ -565,6 +567,7 @@ public interface K extends J {
     @SuppressWarnings({"unused", "unchecked"})
     @FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
     @EqualsAndHashCode(callSuper = false, onlyExplicitlyIncluded = true)
+    @RequiredArgsConstructor
     @AllArgsConstructor(access = AccessLevel.PRIVATE)
     @Data
     final class Constructor implements K, Statement, TypedTree {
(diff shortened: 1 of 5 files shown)


src/main/java/org/openrewrite/kotlin/tree/K.java
```diff
@@ -577,8 +576,28 @@ final class Constructor implements K, Statement, TypedTree {
         Markers markers;
         J.MethodDeclaration methodDeclaration;
         Space colon;
```
knutwannheden, line 578: We should check if we can annotate both `colon` and `constructorInvocation` as `@Nullable` and if possible also as `@JsonIgnore` so that we don't continue serializing them, as we otherwise never will be able to derelease these fields.

knutwannheden: The fields should of course also be annotated as `@Deprecated`.

jkschneider: We don't need to preserve the fields at all. Just need to have a constructor that reallocates their data to the `JLeftPadded`.

knutwannheden: As discussed yesterday we should verify that this also works as expected in a SaaS-like environment (with the corresponding Jackson config).

src/main/java/org/openrewrite/kotlin/tree/K.java
```diff
@@ -567,17 +568,50 @@ public <T extends J> T withType(@Nullable JavaType type) {
...
+
+        // For backward compatibility, handle removed fields `colon` and `constructorInvocation` which has been relocated to `invocation`
+        // Todo, Remove when we feel good that kotlin LSTs have been rebuilt.
+        @Deprecated
+        @JsonCreator
+        public Constructor(UUID id,
+                           Markers markers,
+                           J.MethodDeclaration methodDeclaration,
+                           @JsonProperty("colon") Space colon,
+                           @JsonProperty("constructorInvocation") ConstructorInvocation constructorInvocation) {
```
knutwannheden, line 608: We should declare these two parameters as `@Nullable`. As this constructor is declared as `@JsonConstructor`, Jackson will also call it when deserializing a new LST, which doesn't have these two properties. I tested that with the following tests in `TreeSerializerTest`:
```java
    @Test
    void jsonCreatorTest() throws IOException {
        ObjectMapper mapper = TreeSerializer.builder().build().buildMapper(getClass().getClassLoader(), StreamReadConstraints.defaults());
        ByteArrayOutputStream out = new ByteArrayOutputStream();

        FooOld input = new FooOld("foo");
        mapper.writeValue(out, input);
        FooNew deserialized = mapper.readValue(out.toByteArray(), FooNew.class);
        System.out.println(deserialized);
    }

    @Test
    void jsonCreatorTest2() throws IOException {
        ObjectMapper mapper = TreeSerializer.builder().build().buildMapper(getClass().getClassLoader(), StreamReadConstraints.defaults());
        ByteArrayOutputStream out = new ByteArrayOutputStream();

        FooNew input = new FooNew(2);
        mapper.writeValue(out, input);
        FooNew deserialized = mapper.readValue(out.toByteArray(), FooNew.class);
        System.out.println(deserialized);
    }

    @FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
    @EqualsAndHashCode(callSuper = false, onlyExplicitlyIncluded = true)
    @RequiredArgsConstructor
    @Data
    static class FooOld {
        String name;
    }

    @FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
    @EqualsAndHashCode(callSuper = false, onlyExplicitlyIncluded = true)
    @RequiredArgsConstructor
    @Data
    static class FooNew {
        Integer age;
        @JsonCreator
        public FooNew(String name) {
            this.age = name.length();
        }
    }
```

knutwannheden (approved):

PR #523, opened 2023-12-11
Update `K.Property` model, add `accessors` as a replacement of `setter`/`getter`/`isSetterFirst`

kunli2:
On `K.Property`, the MethodDeclaration already contains the name `get` or `set` such that we don’t need to separate these and have a boolean `isSetterFirst`. Rather, `JContainer<MethodDeclaration> accessors`

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
+++ b/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
@@ -2811,13 +2811,10 @@ public class KotlinTreeParserVisitor extends KtVisitor<J, ExecutionContext> {
         List<J.Modifier> modifiers = mapModifiers(property.getModifierList(), leadingAnnotations, lastAnnotations, data);
         TypeTree typeExpression = null;
         List<JRightPadded<J.VariableDeclarations.NamedVariable>> variables = new ArrayList<>();
-        J.MethodDeclaration getter = null;
-        J.MethodDeclaration setter = null;
         JRightPadded<Expression> receiver = null;
         JContainer<J.TypeParameter> typeParameters = property.getTypeParameterList() != null ?
                 JContainer.build(prefix(property.getTypeParameterList()), mapTypeParameters(property.getTypeParameterList(), data), Markers.EMPTY) : null;
         K.TypeConstraints typeConstraints = null;
-        boolean isSetterFirst = false;
         Set<PsiElement> prefixConsumedSet = preConsumedInfix(property);
 
         modifiers.add(new J.Modifier(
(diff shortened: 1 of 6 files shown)


src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
```diff
@@ -2892,16 +2892,28 @@ public J visitProperty(KtProperty property, ExecutionContext data) {
         }
 
         if (getter != null || setter != null || receiver != null || typeConstraints != null) {
+            List<JRightPadded<J.MethodDeclaration>> rps = new ArrayList<>(2);
+
+            if (setter != null) {
+                rps.add(padRight(setter, Space.EMPTY));
```
knutwannheden, line 2898: This looks a bit strange. Does the right padding represent the space before an optional semi-colon? Looks like `maybeTrailingSemicolon()` should be used here.

src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
```diff
@@ -2892,16 +2892,28 @@ public J visitProperty(KtProperty property, ExecutionContext data) {
...
+            }
+
+            if (getter != null) {
+                rps.add(padRight(getter, Space.EMPTY));
+            }
+
+            if (!isSetterFirst) {
+                Collections.reverse(rps);
```
knutwannheden, line 2906: It seems like it would be simpler to just use `KtProperty#getAccessors()` which should retrieve them in the order they occur in the source code.

src/main/java/org/openrewrite/kotlin/tree/K.java
```diff
@@ -1436,16 +1438,49 @@ public List<TypeParameter> getTypeParameters() {
...
+        public Property(@JsonProperty("id") UUID id,
+                        @JsonProperty("prefix") Space prefix,
+                        @JsonProperty("markers") Markers markers,
+                        @JsonProperty("typeParameters") @Nullable JContainer<TypeParameter> typeParameters,
+                        @JsonProperty("variableDeclarations") VariableDeclarations variableDeclarations,
+                        @JsonProperty("typeConstraints") @Nullable K.TypeConstraints typeConstraints,
+                        @JsonProperty("getter") @Nullable J.MethodDeclaration getter,
+                        @JsonProperty("setter") @Nullable J.MethodDeclaration setter,
+                        @JsonProperty("isSetterFirst") boolean isSetterFirst,
+                        @JsonProperty("receiver") @Nullable JRightPadded<Expression> receiver
```
knutwannheden, line 1459: Also here we should test with a SaaS-like configuration, which shouldn't require these `@JsonProperty` annotations.

src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
kunli2: good catch, The trailing semi-colon of setter/getter is not handled by us. I added a test case for this and fixed.

src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
kunli2: yeah, good point, `KtProperty#getAccessors()` can make it simpler. revised to use `KtProperty#getAccessors()`

src/main/java/org/openrewrite/kotlin/tree/K.java
kunli2: Verified on SAAS and confirmed that `@JsonProperty` is optional. so I removed the redundant `@JsonProperty` for  other fields but kept 3 removed fields to have explicit `@JsonProperty` (it's also optional but I prefer to keep it to emphasize they are removed fields, anyway this method will be removed later after new LSTs are built)

knutwannheden (approved):

Issue #517, opened 2023-12-09
PSI to FIR association issue: Incorrect on parent classes.


traceyyoshima:
```kotlin
                  class A {
                      class B {
                          class C
                      }
                  }
                  val c = A.B.C()
```

Both A and B will have the same type, `A$B`.

kunli2:
This is hard to resolve since `A.B` is the minimum granularity in the FIR, so to distinguish between `A` and `B` will require some additional algorithm on the FIR.

PR #532, opened 2023-12-12
Fix type of nested inner-class field access

kunli2:
fixes https://github.com/openrewrite/rewrite-kotlin/issues/517

### DIFF
--- a/src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
+++ b/src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
@@ -113,10 +113,15 @@ class KotlinTypeMapping(
             val flattened = flattenDots(dots)
             if (flattened != null) {
                 val types = signature.split("$")
-                if (types.size == flattened.size) {
+                if (types.size <= flattened.size) {
                     for ((index, element) in flattened.withIndex()) {
                         if (element == psi) {
-                            signature = types.take(types.size - index).joinToString("$")
+                            signature = if (index < types.size) {
+                                types.take(types.size - index).joinToString("$")
+                            } else {
+                                return null
+                            }
+                            break;
                         }
                     }
                 }
--- a/src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
+++ b/src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
@@ -19,6 +19,7 @@ import org.jetbrains.kotlin.KtFakeSourceElementKind
 import org.jetbrains.kotlin.builtins.PrimitiveType
 import org.jetbrains.kotlin.codegen.classId
 import org.jetbrains.kotlin.codegen.topLevelClassAsmType
+import org.jetbrains.kotlin.com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.descriptors.ClassKind
 import org.jetbrains.kotlin.descriptors.Modality
 import org.jetbrains.kotlin.descriptors.Visibility
@@ -50,6 +51,9 @@ import org.jetbrains.kotlin.load.java.structure.*
 import org.jetbrains.kotlin.load.java.structure.impl.classFiles.*
 import org.jetbrains.kotlin.load.kotlin.JvmPackagePartSource
 import org.jetbrains.kotlin.name.isOneSegmentFQN
+import org.jetbrains.kotlin.psi
+import org.jetbrains.kotlin.psi.KtDotQualifiedExpression
+import org.jetbrains.kotlin.psi.KtNameReferenceExpression
 import org.jetbrains.kotlin.resolve.jvm.JvmClassName
 import org.jetbrains.kotlin.types.ConstantValueKind
 import org.jetbrains.kotlin.types.Variance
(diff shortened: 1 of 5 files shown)


src/test/java/org/openrewrite/kotlin/tree/FieldAccessTest.java
```diff
@@ -249,4 +253,51 @@ inner class B {
           )
         );
     }
+
+    @Issue("https://github.com/openrewrite/rewrite-kotlin/issues/517")
+    @Test
+    void nestedFieldAccessType() {
```
traceyyoshima, line 259: The tree tests are meant for idempotent prints, and type mapping tests exist in KotlinTypeMapping.
Please move this to KotlinTypeMapping.

Note:
This is a revision to the test that doesn't involve transformations.
Notice that A, B, and C are each Identifiers, where types are being fixed.
That allows visitIdentifier to check each case.
The test adds a package to ensure that imports, which are also field access selects, do not contain types (I didn't add those assertions, but the types look good!)

```java
        @Issue("https://github.com/openrewrite/rewrite-kotlin/issues/517")
        @Test
        void nestedFieldAccessType() {
            rewriteRun(
              kotlin(
                """
                  package foo.bar
                  
                  class A {
                      class B {
                          class A {
                              class C
                          }
                      }
                  }
                  """
              ),
              kotlin(
                """
                  import foo.bar.A
                  
                  val x = A.B.A.C()
                  """,
                spec -> spec.afterRecipe(cu -> {
                      AtomicBoolean a1 = new AtomicBoolean(false);
                      AtomicBoolean a2 = new AtomicBoolean(false);
                      AtomicBoolean b = new AtomicBoolean(false);
                      AtomicBoolean c = new AtomicBoolean(false);
                          @Override
                          public J.Identifier visitIdentifier(J.Identifier identifier, Integer n) {
                              switch (identifier.getSimpleName()) {
                                  case "A": {
                                      if (identifier.getType() != null) {
                                          if ("".equals(identifier.getType().toString())) {
                                              a1.set(true);
                                          } else if ("".equals(identifier.getType().toString())) {
                                              a2.set(true);
                                          }
                                      }
                                  }
                                  case "B": {
                                      assertThat(identifier.getType().toString()).isEqualTo("");
                                      b.set(true);
                                  }
                                  case "C": {
                                      assertThat(identifier.getType().toString()).isEqualTo("");
                                      c.set(true);
                                  }
                              }
                              return super.visitIdentifier(identifier, n);
                          }
                      }.visit(cu, 0);
                      assertThat(a1.get()).isTrue();
                      assertThat(a2.get()).isTrue();
                      assertThat(b.get()).isTrue();
                      assertThat(c.get()).isTrue();
                  }
                )
              )
            );
        }
```

src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
```diff
@@ -98,6 +102,54 @@ class KotlinTypeMapping(
         return type(type, parent, signature)
     }
 
+    fun type(psi: PsiElement?,  type: Any?, parent: Any?): JavaType? {
```
traceyyoshima, line 105: The new type method seems to blend the functionality of the `PsiElementAssociations`, `KotlinSignatureBuilder`, and `KotlinTypeMapping`.

What do you think about detecting the case in the `PsiElementAssocation` which is meant to associate the PSI to the FIR? And have a new method in the signature builder that is capable of returning the expected signature. Then the TypeMapping may use the corrected signature to return the correct type?

Note:
If I understand the changes correctly, it will fail for `val x = foo.bar.A.B.A.C()`. Please test that as well.

src/test/java/org/openrewrite/kotlin/tree/FieldAccessTest.java
kunli2: talked offline, will keep it as is.

traceyyoshima (changes requested):
confirmed the solution does not work for: `val x = foo.bar.A.B.A.C()`

kunli2:
> confirmed the solution does not work for: `val x = foo.bar.A.B.A.C()`

fixed by the 2nd commit, and added the test case as well.

src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
kunli2: I introduced this method to be able to break down a leaf FIR node (like `A.B.A` in this case) into pieces (like breaking down the atomic nucleus into quarks), this requires PSI's help, so that's why it has a psi as a parameter.
I didn't find another way to break down a leaf FIR node like `A.B.A` to `A`, `B`, and `A`.

traceyyoshima: I mean to apply the same concept, but in a different way that encapsulates the functionality in the applicable classes. I'll take a look tomorrow

traceyyoshima: I took a look at the changes, and there are still issues. The changes cause parameterization to be lost since the signature created by the signature build is based on flattening a dot-qualified expression rather than using the FIR type attribution.

I tested the original suggestion about using the `ClassId` and `FirSession` to associate the PSI to the FIR and confirmed it's possible. So, we can detect the mismatch in the PsiElementAssociation and find the right FIR element, then call `typeMapping#type(...)` on the associated FIR.

To recap `ClassId` and `FirSession`:
A `FirResolvedQualifier` will have a symbol if it is resolved to a type.
The symbol will have access to a `ClassId,` which enables the detection of outer classes.
Then, `ClassId` can be used to look up the associated FIR element, which we can use for type attribution.

I think we can close this PR since it still isn't quite right and it'll be safer to use the FirSession instead.

kunli2:
An alternate PR : https://github.com/openrewrite/rewrite-kotlin/pull/547/files

PR #534, opened 2023-12-12
Set constructor type on enum values.

### LINKED ISSUE #533: PSI association issue: parser visitor sets enum constructors to null
```kotlin
              enum class Code {
                  YES ,
              }
              enum class Test ( val arg: Code , ) {
                  FOO ( Code.YES , ) {
                      // Body is required to reproduce issue
                  }
              }
```

traceyyoshima:
EnumValues will contain the constructor type instead of null.
fixes #533

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
+++ b/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
@@ -494,14 +494,13 @@ public class KotlinTreeParserVisitor extends KtVisitor<J, ExecutionContext> {
 
         Set<PsiElement> consumedSpaces = preConsumedInfix(enumEntry);
 
-        // TODO: in java the EnumValue has a type of JavaType.Variable with a null fieldType.
         J.Identifier name = createIdentifier(requireNonNull(enumEntry.getNameIdentifier()), type(enumEntry), consumedSpaces);
         J.NewClass initializer = null;
 
+        JavaType.Method mt = methodDeclarationType(enumEntry);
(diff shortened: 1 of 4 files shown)


src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
```diff
@@ -494,14 +494,12 @@ public J visitEnumEntry(KtEnumEntry enumEntry, ExecutionContext data) {
...
 
-        // TODO: in java the EnumValue has a type of JavaType.Variable with a null fieldType.
         J.Identifier name = createIdentifier(requireNonNull(enumEntry.getNameIdentifier()), type(enumEntry), consumedSpaces);
         J.NewClass initializer = null;
 
         if (enumEntry.getInitializerList() != null) {
-            // TODO: this creates an empty init with no type attribution: enum class Foo { BAR() }
-            //   Add constructor method type.
             initializer = (J.NewClass) enumEntry.getInitializerList().accept(this, data);
+            initializer = initializer.withMethodType(methodDeclarationType(enumEntry));
```
kunli2, line 502: nit: we can have a variable here for `methodDeclarationType(enumEntry)` to save a method invocation.

kunli2 (approved):

src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
traceyyoshima: I'll move the type up into a variable so it's added in both positions

PR #542, opened 2023-12-13
Introduce `K.UseSite` and deprecate `AnnotationUseSite` marker

kunli2:
1. Introduce `K.UseSite` which is a `NameTree`, and will be the `annotationType` in a `J.Annotation`.
2. Deprecate `AnnotationUseSite` marker
3. For empty name use-site annotation like this `@[ Anno1 Anno2 ]`, the target will be an `J.Empty`

Notes:
This PR requires a small rewrite code change to `J.Annotation#getSimpleName()`
from
```java
        public String getSimpleName() {
            return annotationType instanceof Identifier ?
                    ((Identifier) annotationType).getSimpleName() :
                    ((J.FieldAccess) annotationType).getSimpleName();
        }
```
to
```java
        public String getSimpleName() {
            if (annotationType instanceof Identifier) {
                return ((Identifier) annotationType).getSimpleName();
            } else if (annotationType instanceof J.FieldAccess ) {
                return ((J.FieldAccess) annotationType).getSimpleName();
            } else {
                return annotationType.toString();
            }
        }
```

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/internal/KotlinPrinter.java
+++ b/src/main/java/org/openrewrite/kotlin/internal/KotlinPrinter.java
@@ -433,6 +433,19 @@ public class KotlinPrinter<P> extends KotlinVisitor<PrintOutputCapture<P>> {
     }
 
     @Override
+    public J visitUseSite(K.UseSite useSite, PrintOutputCapture<P> p) {
+        beforeSyntax(useSite, Space.Location.ANNOTATIONS, p);
+        visit(useSite.getTarget(), p);
+        visitSpace(useSite.getPadding().getTarget().getAfter(), Space.Location.ANNOTATIONS, p);
+
+        if (!(useSite.getTarget() instanceof J.Empty)) {
+            p.append(":");
+        }
+
+        return useSite;
+    }
+
+    @Override
     public J visitWhen(K.When when, PrintOutputCapture<P> p) {
         beforeSyntax(when, KSpace.Location.WHEN_PREFIX, p);
         p.append("when");
@@ -484,12 +497,11 @@ public class KotlinPrinter<P> extends KotlinVisitor<PrintOutputCapture<P>> {
             String afterArgs = ")";
             String delimiter = ",";
 
-            AnnotationUseSite useSite = annotation.getMarkers().findFirst(AnnotationUseSite.class).orElse(null);
-            if (useSite != null) {
-                kotlinPrinter.visitSpace(useSite.getPrefix(), KSpace.Location.ANNOTATION_CALL_SITE_PREFIX, p);
-                p.append(":");
+            boolean isUseSite = annotation.getAnnotationType() instanceof K.UseSite;
+            if (isUseSite) {
+                boolean isImplicitBracket = ((K.UseSite) annotation.getAnnotationType()).isImplicitBracket();
 
-                if (!useSite.isImplicitBracket()) {
+                if (!isImplicitBracket) {
                     beforeArgs = "[";
                     afterArgs = "]";
                 } else {
--- a/src/main/java/org/openrewrite/kotlin/internal/KotlinPrinter.java
+++ b/src/main/java/org/openrewrite/kotlin/internal/KotlinPrinter.java
@@ -434,7 +434,7 @@ public class KotlinPrinter<P> extends KotlinVisitor<PrintOutputCapture<P>> {
 
     @Override
     public J visitUseSite(K.UseSite useSite, PrintOutputCapture<P> p) {
-        beforeSyntax(useSite, Space.Location.ANNOTATIONS, p);
+        beforeSyntax(useSite, Space.Location.ANNOTATION_PREFIX, p);
         visit(useSite.getTarget(), p);
         visitSpace(useSite.getPadding().getTarget().getAfter(), Space.Location.ANNOTATIONS, p);
 
--- a/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
+++ b/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
@@ -1675,9 +1675,14 @@ public class KotlinTreeParserVisitor extends KtVisitor<J, ExecutionContext> {
 
     @Override
     public J visitAnnotation(KtAnnotation annotation, ExecutionContext data) {
-        if (annotation.getUseSiteTarget() == null) {
-            throw new UnsupportedOperationException("TODO, Some cases we don't know");
+        NameTree target;
+
+        if (annotation.getUseSiteTarget() != null) {
+            target = (NameTree) annotation.getUseSiteTarget().accept(this, data);
+        } else {
+            target = new J.Empty(randomId(), Space.EMPTY, Markers.EMPTY);
         }
+
         List<KtAnnotationEntry> annotationEntries = annotation.getEntries();
         List<JRightPadded<Expression>> rpAnnotations = new ArrayList<>(annotationEntries.size());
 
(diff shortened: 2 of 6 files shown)


src/main/java/org/openrewrite/kotlin/internal/KotlinPrinter.java
```diff
@@ -432,6 +432,19 @@ public J visitUnary(K.Unary unary, PrintOutputCapture<P> p) {
         return unary;
     }
 
+    @Override
+    public J visitUseSite(K.UseSite useSite, PrintOutputCapture<P> p) {
+        beforeSyntax(useSite, Space.Location.ANNOTATIONS, p);
```
knutwannheden, line 437: Not quite sure, but I suspect it should be like this:
```suggestion
        beforeSyntax(useSite, Space.Location.ANNOTATION_PREFIX, p);
```

src/main/java/org/openrewrite/kotlin/internal/KotlinPrinter.java
```diff
@@ -432,6 +432,19 @@ public J visitUnary(K.Unary unary, PrintOutputCapture<P> p) {
...
+        visit(useSite.getTarget(), p);
+        visitSpace(useSite.getPadding().getTarget().getAfter(), Space.Location.ANNOTATIONS, p);
```
knutwannheden, line 439: I would have expected some kind of SUFFIX space location here. Probably specific to K.UseSite.

src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
```diff
@@ -1698,8 +1703,8 @@ public J visitAnnotation(KtAnnotation annotation, ExecutionContext data) {
 
         return mapType(new J.Annotation(randomId(),
                 Space.EMPTY,
-                Markers.EMPTY.addIfAbsent(new AnnotationUseSite(randomId(), suffix(annotation.getUseSiteTarget()), isImplicitBracket)),
-                (NameTree) annotation.getUseSiteTarget().accept(this, data),
+                Markers.EMPTY,
+                new K.UseSite(randomId(), Space.EMPTY, Markers.EMPTY, padRight(target, suffix(annotation.getUseSiteTarget())), isImplicitBracket),
```
knutwannheden, line 1707: Do I understand it correctly that now every annotation will have a `K.UseSite` and that this will wrap a `J.Empty` when there wasn't any use-site target? That would be wrong, I think.

knutwannheden (changes requested):
@kunli2 I think it would really help if we could draw object diagrams showing what the LST model looks like for the various cases like:
- `@Ann`
- `@field:Ann`
- `@[Ann1 Ann2]`
- `@field:[Ann1 Ann2]`

src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
kunli2: Not every annotation will have `K.UseSite`, but just use-site annotations which include below 4 cases
```
@[foo, bar]     // This will have a `J.Empty` as the target
@field: foo
@field: [ foo ]
@field: [ foo bar]
```

kunli2:
> @kunli2 I think it would really help if we could draw object diagrams showing what the LST model looks like for the various cases like:
> 
> * `@Ann`
> * `@field:Ann`
> * `@[Ann1 Ann2]`
> * `@field:[Ann1 Ann2]`

Sure, here are all the cases, 1 regular annotation 4 use-site annotations, and the LST structure.
`K.UseSite` will be the `annotationType` of the `J.Annotation`, and the following annotations will be parameters
```

// case 0, Non use-site, regular annotation
@Ann1

\---J.Annotation | "@Ann1"
    \---J.Identifier | "Ann1"


// case 1, use-site, implicit bracket
@field : Ann1

\---J.Annotation
    |----J.Annotation | "@Ann1"    // Parameter
    |  \---J.Identifier | "Ann1"
    \---K.UseSite | isImplicitBracket = true 
       \-------J.Identifier | "field"

// case 2, use-site, explicit bracket
@field : [ Ann1 ]

\---J.Annotation
    |---J.Annotation | "@Ann1"      // Parameter
    |   \---J.Identifier | "Ann1"
    \---K.UseSite | isImplicitBracket = false 
        \-------J.Identifier | "field"

// case 3, use-site, multi annotations with explicit bracket
@field : [Ann1 Ann2]

\---J.Annotation
    |-----------J.Annotation | "@Ann1"      // Parameter 1
    |   |       \---J.Identifier | "Ann1"
    |   \-------J.Annotation | "@Ann2"      // Parameter 2
    |           \---J.Identifier | "Ann2"
    \---K.UseSite | isImplicitBracket = false
        \-------J.Identifier | "field"


// case 4, use-site, empty target
@[ Ann1 Ann2]

\---J.Annotation
    |-----------J.Annotation | "@Ann1"      // Parameter 1
    |   |       \---J.Identifier | "Ann1"
    |   \-------J.Annotation | "@Ann2"      // Parameter 2
    |           \---J.Identifier | "Ann2"
    \---K.UseSite | isImplicitBracket = false
        \-------J.Empty

```

src/main/java/org/openrewrite/kotlin/internal/KotlinPrinter.java
kunli2: I think the `JRightPadded<NameTree> target` is enough here, as the right padding is the space before the colon.
```kotlin
@field /*Right padding of the target*/ : Ann1
```

src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
kunli2: the regular annotation is still a `J.Annotation` with no change and no `K.UseSite` there, `K.UseSite` is for use-site annotations only

knutwannheden:
Case 1 seems weird to me. Can we come up with a scheme where UseSite owns both `field` and `Ann1` and there is only one `J.Annotation`? Having `Ann1` as a parameter to some dummy annotation seems very strange.

kunli2:
> Case 1 seems weird to me. Can we come up with a scheme where UseSite owns both `field` and `Ann1` and there is only one `J.Annotation`? Having `Ann1` as a parameter to some dummy annotation seems very strange.

If I understand this correctly, you do want to put both  `field` and `Ann1` to `K.UseSite`, and  `K.UseSite` is still an `annotationType` field in a `J.Annotation`, and the `J.Annotation` will have empty parameters, if so, I think it's doable, and seems this makes sense too.

knutwannheden:
> please confirm so I can come up with a new PR with this approach.

Yes, that is what I meant. For the multi-annotation case we will probably still require a "dummy" annotation.

Possibly something along these lines (pseudocode):

```java
public interface K extends J {

  final class AnnotationType implements K, NameTree, Expression {
    Space prefix;
    JRightPadded<J.Identifier> useSite;
    TypeTree typeTree;

    public JavaType getType() {
      return typeTree.getType();
    }
  }

  final class MultiAnnotationType implements K, NameTree, Expression {
    Space prefix;
    @Nullable
    JRightPadded<J.Identifier> useSite;

    public JavaType getType() {
      return null;
    }
  }
}
```

- `K.AnnotationType` would be used for `J.Annotation#annotationType` whenever there is a single annotation with a use-site target
- `K.MultiAnnotationType` would be used for all multi-annotations (regardless of whether there is a use-site target or not)
- If the multi annotations then end up as nested inside the `MultiAnnotationType` or in the parameters of `J.Annotation` needs to be checked

@kunli2 What do you think?

kunli2:
we composed this [doc](https://docs.google.com/document/d/1vV9I-yTpVj7lyDxO7HVpO_x7lOyrCQIGEgQ-sCHqoRA/edit?usp=sharing) for LST model proposals. discussed with the team and decided to go option 3. so closing out this PR and I will have a new PR with option 3

Issue #482, opened 2023-11-30
Info to extend TypeValidation for Kotlin


traceyyoshima:
#### Code
```kotlin
@file:Suppress("PLATFORM_CLASS_MAPPED_TO_KOTLIN", "unused")
```

#### Info on use-site name
A use-site annotation is a keyword that informs the compiler where to apply annotation.
The use-site annotation does not resolve to an annotation with a type, which means the type is null.
FindMissingTypes does not consider this case as a valid null type, which leads to an error.

#### Code
```kotlin
import java.util.regex.Pattern.CASE_INSENSITIVE as i
class A {
    val f = arrayOf(i)
}
```
#### Info
The variable type of a type aliases identifier is constructed from the aliased type. So, the aliased type and the J.Identifier's name do not match. The identifier `i` has a type of `java.util.regex.Pattern{name=CASE_INSENSITIVE,type=kotlin.Int}`

#### Code
```kotlin
val a = mapOf ( 1 to "one" , 2 to "two" )
val b = a [ 1 ]
```

#### Info on `get` and `set`
Kotlin recommends using `map[key]` instead of `map.get(key)`. The type associated with the `get` is a MethodType with the name `get`, which does not match the name of the method invocation created `a`. A similar issue happens with `set`.

#### Code
```kotlin
              abstract class Test ( arg : Test  .   ( )  ->   Unit    =  { } ) {
                  init {
                      arg ( )
                  }
              }
```
#### Info on `invoke`
A similar issue exists for implicit invoke calls.

#### Code
```kotlin
val positive = fun(i: Int) = i > 0 /* C */
```
#### Info on `anonymous functions`
Anonymous functions have the name `anonymous`, which does not match the identifier created with an empty name.

#### Code
```kotlin
              operator fun <T> MutableCollection<T>.divAssign(elements: Collection<T>) {
                  this.retainAll(elements)
              }
```
#### Info on receivers
An extension receiver has an unused name, `<receiverType>`, which does not match the variable type name.

#### Code
```kotlin
              import kotlin.text.Regex

              fun foo(choices: List<Pair<String, String>>, peekedHeader: Regex) {
                  for ((_, adapter) in choices) {
                      if (adapter.matches(peekedHeader)) {
                          print("1")
                      }
                  }
              }
```
#### Info on destructuring declarations
Similar to receivers, a destruct is created with a name `<destruct>`

### MemberReference
Null types on class literal expressions `Type::class`.
[...]

Issue #474, opened 2023-11-28
Add type validation to Assertions


traceyyoshima:
In Java, type validation helps to identify type issues when recipes make unexpected changes. Similarly, Kotlin will benefit from implementing type validation to inform users of changes that may lead to unexpected results.

Kotlin-type validation requires a customized visitor compared to java to handle language differences. The differences will be documented and discussed [here](https://github.com/openrewrite/rewrite-kotlin/issues/482).

Note: requires the ability to turn off type validation on sources that intentionally do not resolve types

Issue #480, opened 2023-11-30
Keyword else should not contain a type


traceyyoshima:
```kotlin
fun method ( i : Int ) : String {
    when {
        i . mod ( 2 ) == 0 -> return "even"
        else -> return "odd"
    }
}
```

traceyyoshima:
The type validation issue is caused by the else keyword being parsed as an Identifier through `visitSimpleName`.
The PSI cannot distinguish between keywords and resolvable types, so `type` is called on everything.

The result is a keyword with a type, which is incorrect and leads to a validation error.

traceyyoshima:
Closing issue: this will be added as acceptable in type validation

PR #543, opened 2023-12-13
Added type validation to Assertions.

traceyyoshima:
Changes:
- Kotlin test harness will validate types.

fixes https://github.com/openrewrite/rewrite-kotlin/issues/482
fixes https://github.com/openrewrite/rewrite-kotlin/issues/474

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/Assertions.java
+++ b/src/main/java/org/openrewrite/kotlin/Assertions.java
@@ -21,30 +21,43 @@ import org.jetbrains.annotations.NotNull;
 import org.openrewrite.*;
 import org.openrewrite.internal.ThrowingConsumer;
 import org.openrewrite.internal.lang.Nullable;
-import org.openrewrite.java.tree.J;
-import org.openrewrite.java.tree.Space;
-import org.openrewrite.java.tree.TextComment;
+import org.openrewrite.java.search.FindMissingTypes;
+import org.openrewrite.java.tree.*;
+import org.openrewrite.kotlin.marker.AnnotationUseSite;
+import org.openrewrite.kotlin.marker.Extension;
+import org.openrewrite.kotlin.marker.IndexedAccess;
 import org.openrewrite.kotlin.tree.K;
 import org.openrewrite.kotlin.tree.KSpace;
+import org.openrewrite.marker.Marker;
 import org.openrewrite.marker.Markers;
+import org.openrewrite.marker.SearchResult;
 import org.openrewrite.test.SourceSpec;
 import org.openrewrite.test.SourceSpecs;
+import org.openrewrite.test.TypeValidation;
 import org.openrewrite.test.UncheckedConsumer;
 import org.opentest4j.AssertionFailedError;
 
-import java.util.Collections;
-import java.util.Optional;
+import java.util.*;
 import java.util.function.Consumer;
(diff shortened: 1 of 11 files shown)


traceyyoshima:
Blocked by https://github.com/openrewrite/rewrite-kotlin/issues/480.

src/main/java/org/openrewrite/kotlin/Assertions.java
```diff
@@ -196,6 +209,36 @@ public static UncheckedConsumer<SourceSpec<?>> spaceConscious() {
         };
     }
 
+    private static void assertValidTypes(TypeValidation typeValidation, J sf) {
```
traceyyoshima, line 212: @knutwannheden @kunli2, I don't think mixins will help us customize the `FindMissingTypes` recipe. Initially, I thought about creating a `FindMissingKotlinTypes` recipe/visitor, which could arguably be helpful. But I recall Jonathan preferring nearly identical recipes not to be created. Eventually, if we find there is a need for the augmented recipe, we can move the code.

What do you think? Is it fine to have the visitor in Assertions for now or would you prefer a recipe?

src/test/java/org/openrewrite/kotlin/tree/AnnotationTest.java
```diff
@@ -15,48 +15,23 @@
...
+import org.openrewrite.test.TypeValidation;
 
 import java.util.List;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.openrewrite.kotlin.Assertions.kotlin;
 
-@SuppressWarnings({"RedundantSuppression", "RedundantNullableReturnType", "RedundantVisibilityModifier", "UnusedReceiverParameter", "SortModifiers", "TrailingComma"})
+@SuppressWarnings({"RedundantSuppression", "RedundantNullableReturnType", "RedundantVisibilityModifier", "UnusedReceiverParameter", "SortModifiers", "TrailingComma", "RedundantGetter", "RedundantSetter"})
 class AnnotationTest implements RewriteTest {
```
traceyyoshima, line 33: I cleaned up the errors/warnings here when I looked through the tests.

traceyyoshima:
@knutwannheden type validation is essentially ready for review. The failing tests are identifiers with aliases, and I'm looking into how to detect aliases.

```kotlin
import java.util.regex.Pattern.CASE_INSENSITIVE as i

class A {
    val f = arrayOf(i)
}
```

src/main/java/org/openrewrite/kotlin/Assertions.java
knutwannheden: At some point I would be interested in the details regarding the mixins, do that we can try to come up with a fix or another approach for this. An alternative might be delegation kind of like we do in the printer.

For now I think what you currently have is good. Having type validations working is really a step forward. Thanks for fixing this!

knutwannheden (approved):
As we can disable validations on a test case basis, we can fix any remaining issues in follow up PRs.

traceyyoshima:
> As we can disable validations on a test case basis, we can fix any remaining issues in follow up PRs.

Opened: https://github.com/openrewrite/rewrite-kotlin/issues/545 to track a fix for aliases. Tracking the declarations of a name will allow us to apply the appropriate field type comparison.

PR #547, opened 2023-12-14
Associate the inner classes to the PSI.

traceyyoshima:
Changes:

- Nested classes will have the correct type attribution.

fixes #517

PR #548, opened 2023-12-15
Improve condition of nested inner class type fetching

kunli2:
To address the comment in this [PR](https://github.com/openrewrite/rewrite-kotlin/pull/547).
I'm concerned that the conditions here may not be very persuasive and conducive to understanding, so there should be an alternative condition of using PSI/FIR elements.

so instead of `parts.size == 2 && psi.text == parts[0]`, I think the condition can be optimized to `psi !is KtDotQualifiedExpression && psi == parent.receiverExpression`


traceyyoshima (approved):
LGTM!

PR #549, opened 2023-12-16
New LST models to support use-site annotation and deprecate `AnnotationUseSite` marker

kunli2:
After discussion the design [doc](https://docs.google.com/document/d/1vV9I-yTpVj7lyDxO7HVpO_x7lOyrCQIGEgQ-sCHqoRA/edit?usp=sharing) with the team, we decide to use option 3 to support use-site annotation instead of the `AnnotationUseSite` marker.

Here are more details of option 3 
Two models for single and multiple use-site included annotations, and both of them will be an `annotationType` field of the `J.Annotation`.

Schema:
```java
class AnnotationType implements K, NameTree {
    Space prefix;
    @Nullable JRightPadded<Expression> useSite;
    J.Annotation callee;
  }

 class MultiAnnotationType implements K, NameTree {
    Space prefix;
    @Nullable JRightPadded<Expression> useSite;
    JContainer<J.Annotation> annotations;
  }

```

Use case examples:

1. Single use-site annotation

```
@field : Ann1
-----------------------
"@field : Ann1" is a J.Annotation
	annotationType = K.AnnotationType
        arguments = []
"@field : Ann1" is a K.AnnotationType
	useSite = field + rightPadding
        callee = Ann1
```

2. Single annotation in bracket
```
@field : [ Ann1 ]
-----------------------
"@field : [ Ann1 ]": J.Annotation
	annotationType = K.MultiAnnotationType
        arguments = []
"@field : [ Ann1 ]": K.MultiAnnotationType, 
        useSite = field + rightPadding
        annotations = [Ann1]
```

3. Multi annotations
```
@field : [Ann1 Ann2]
-----------------------
"@field : [Ann1 Ann2]": J.Annotation
	annotationType = K.MultiAnnotationType
        arguments = []
"@field : [Ann1 Ann2]": K.MultiAnnotationType
	useSite = field + rightPadding
        annotations = [Ann1, Ann2]

```
4. Multi annotations without target
```
@[ Ann1 Ann2]
-----------------------
"@[ Ann1 Ann2]": J.Annotation
	annotationType = K.MultiAnnotationType
        arguments = []
"@[ Ann1 Ann2]": K.MultiAnnotationType: 
        useSite = J.Empty
        annotations = [Ann1, Ann2]
```

Notes:
This PR requires a small rewrite code change to `J.Annotation#getSimpleName()`
from
```java
        public String getSimpleName() {
            return annotationType instanceof Identifier ?
                    ((Identifier) annotationType).getSimpleName() :
                    ((J.FieldAccess) annotationType).getSimpleName();
        }
```
to
```java
        public String getSimpleName() {
            if (annotationType instanceof Identifier) {
                return ((Identifier) annotationType).getSimpleName();
            } else if (annotationType instanceof J.FieldAccess ) {
                return ((J.FieldAccess) annotationType).getSimpleName();
            } else {
                return annotationType.toString();
            }
        }
```


src/main/java/org/openrewrite/kotlin/internal/KotlinPrinter.java
```diff
@@ -417,6 +417,28 @@ public J visitUnary(K.Unary unary, PrintOutputCapture<P> p) {
...
+        visit(annotationType.getCallee(), p);
+        return annotationType;
+    }
+
+    @Override
+    public J visitMultiAnnotationType(K.MultiAnnotationType multiAnnotationType, PrintOutputCapture<P> p) {
+        beforeSyntax(multiAnnotationType, Space.Location.ANNOTATION_PREFIX, p);
+        visitRightPadded(multiAnnotationType.getPadding().getUseSite(), p);
+
+        if (!(multiAnnotationType.getUseSite() instanceof J.Empty)) {
```
knutwannheden, line 434: A `J.Empty` seems a bit odd. Why not simply `null`?

knutwannheden:
@kunli2 AFAICT this will still require us to add a `getSimpleName()` method to `NameTree`. I will create a PR for this and then ask Jonathan for his input on that.

knutwannheden:
> @kunli2 AFAICT this will still require us to add a `getSimpleName()` method to `NameTree`. I will create a PR for this and then ask Jonathan for his input on that.

Turns out we don't need that quite yet after all. We had a call site in `KotlinPrinter` which called `J.Annotation#getSimpleName()` for all annotations: https://github.com/openrewrite/rewrite-kotlin/pull/549/commits/12f31a05ab935d5e95b89c905f60d7333d253f23. By removing that the tests now pass.

Since recipes should be allowed to call this method, we should still address this problem. In a first step we could maybe just return the empty string, if the annotation type isn't a `J.Identifier` or `J.FieldAccess`.

knutwannheden:
I've now gone ahead and made the changes to `J.Annotation#getSimpleName()`: https://github.com/openrewrite/rewrite/commit/f7772889cb6e28427bf1ac78566fd6df45f47ba5

PR #551, opened 2023-12-19
Added K.TypeAlias.

### LINKED ISSUE #550: Add new `K.TypeAlias` element
Type aliases are currently modelled as variable declarations with a special `typealias` modifier on them. While this works syntactically, it doesn't make much sense from a model perspective, as a `typealias` doesn't really introduce a new variable and as such would end up confusing a lot of existing recipes. Instead, there should be a dedicated `K.TypeAlias` model element for this.

traceyyoshima:
Changes:
- typealias is a `K.TypeAlias` instead of a `J.VariableDeclaration`.
- Added missing visits in `KotlinIsoVisitor`.
- Updated `RenameTypeAlias` recipe.

fixes #550


src/main/java/org/openrewrite/kotlin/KotlinVisitor.java
```diff
@@ -294,8 +312,35 @@ public J visitSpreadArgument(K.SpreadArgument spreadArgument, P p) {
         return s;
     }
 
+    public J visitTypeAlias(K.TypeAlias typeAlias, P p) {
+        K.TypeAlias t = typeAlias;
+        t = t.withPrefix(visitSpace(t.getPrefix(), KSpace.Location.TYPE_CONSTRAINT_PREFIX, p));
```
knutwannheden, line 317: `TYPE_ALIAS_PREFIX`?

src/main/java/org/openrewrite/kotlin/KotlinVisitor.java
```diff
@@ -294,8 +312,35 @@ public J visitSpreadArgument(K.SpreadArgument spreadArgument, P p) {
...
+        t = t.withModifiers(ListUtils.map(t.getModifiers(),
+                mod -> mod.withPrefix(visitSpace(mod.getPrefix(), Space.Location.MODIFIER_PREFIX, p))));
+        t = t.withModifiers(ListUtils.map(t.getModifiers(), m -> visitAndCast(m, p)));
+        t = t.withName(visitAndCast(t.getName(), p));
+        if (t.getPadding().getTypeParameters() != null) {
+            t = t.getPadding().withTypeParameters(visitContainer(t.getPadding().getTypeParameters(), JContainer.Location.TYPE_PARAMETERS, p));
+        }
+        if (t.getPadding().getInitializer() != null) {
+            t = t.getPadding().withInitializer(visitLeftPadded(t.getPadding().getInitializer(),
+                    JLeftPadded.Location.VARIABLE_INITIALIZER, p));
```
knutwannheden, line 335: Should we add a dedicated space location?

src/main/java/org/openrewrite/kotlin/KotlinVisitor.java
```diff
@@ -331,8 +328,7 @@ public J visitTypeAlias(K.TypeAlias typeAlias, P p) {
             t = t.getPadding().withTypeParameters(visitContainer(t.getPadding().getTypeParameters(), JContainer.Location.TYPE_PARAMETERS, p));
         }
         if (t.getPadding().getInitializer() != null) {
-            t = t.getPadding().withInitializer(visitLeftPadded(t.getPadding().getInitializer(),
-                    JLeftPadded.Location.VARIABLE_INITIALIZER, p));
+            t = t.getPadding().withInitializer(visitLeftPadded(t.getPadding().getInitializer(), p));
```
knutwannheden, line 331: Possibly we can improve the pattern with the parallel structure of `JLeftPadded` vs `KLeftPadded` etc. but as long as we have it, I think we should have corresponding overloads for e.g. `visitLeftPadded()` and use those here in the visitor. Can you add an overload for `KLeftPadded`. It will probably require copying some of the code for the `JLeftPadded` overload.

src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
```diff
@@ -1300,48 +1300,35 @@ public J visitTypeAlias(KtTypeAlias typeAlias, ExecutionContext data) {
         List<J.Annotation> lastAnnotations = new ArrayList<>();
         Set<PsiElement> consumedSpaces = preConsumedInfix(typeAlias);
 
-        List<J.Modifier> modifiers = new ArrayList<>(mapModifiers(typeAlias.getModifierList(), leadingAnnotations, lastAnnotations, data));
+        List<J.Modifier> modifiers = mapModifiers(typeAlias.getModifierList(), leadingAnnotations, lastAnnotations, data);
         modifiers.add(new J.Modifier(randomId(), prefix(typeAlias.getTypeAliasKeyword(), consumedSpaces), markers, "typealias", J.Modifier.Type.LanguageExtension, emptyList()));
```
knutwannheden, line 1304: If we don't need this modifier anymore, I think we should remove it. At the same time we should keep the code which is there to support type aliases created this way (for backwards compatibility). But we should also add a comment to these code blocks, so that nobody inadvertently removes them.

traceyyoshima: I'm fine with removing the modifier -- it's mostly to preserve whitespace/comments. It seems odd to have either a `JLeftPadded<Name> name` or a `Space beforeTypeAlias` to support `internal typealias X = Y`, so I opted to keep the modifier in place.

traceyyoshima:
@knutwannheden do you think this is ready to merge, since any further improvements would be additive?

knutwannheden (approved):

Issue #576, opened 2023-12-27
Issue with change type on an annotation with a use site target


rogerpeckham-toast:
## What version of OpenRewrite are you using?

I am using

- Gradle plugin v6.6.3
- rewrite-kotlin v1.8.4

## How are you running OpenRewrite?

I am using the Gradle plugin to run ChangeType recipes.

## What is the smallest, simplest way to reproduce the problem?

I have created the test [here](https://github.com/rogerpeckham-toast/rewrite-kotlin/blob/change_type_issue/src/test/java/org/openrewrite/kotlin/ChangeTypeTest.java#L488C9-L488C9). The test is a change type recipe for an annotation class that is used with a use site target. When run the import is not added for the new `Target` annotation. 

After stepping through the code I am starting to believe that for some reason the `AddImport` visitor is not actually causing the issue because when the type is changed the underlying type in use isn't changed. To explain using the annotations from the linked test, it doesn't add an import for the `Target` annotation because it thinks that the `Target` annotation added actually has the type `a.b.Original`.

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?

If I figure out the source of this issue I might be able to fix it but there is a good chance I will not be able to fix it.

knutwannheden:
This has now been fixed and will be included in the next release. You can expect a new release within the next two weeks. In the mean time you can use snapshot builds, which should already be available.

PR #577, opened 2023-12-27
Add test for change type of annotation with use site target

rogerpeckham-toast:
## What's changed?
This adds a test for issue #576


knutwannheden:
@rogerpeckham-toast Thanks a lot for the issue and the test case. I've added a commit with a fix in it. There were some recent changes to better support annotations with use-site targets and it turned out that there was something missing there.

PR #581, opened 2023-12-29
Support mod-assign operator `%=`

### LINKED ISSUE #555: Parsing issue: UnsupportedOperation on `%=`
```kotlin
fun method ( n1: Int, n2: Int ) {
    var copy = n1
    copy %= n2
}
```

kunli2:
fixes #555

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
+++ b/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
@@ -1898,15 +1898,17 @@ public class KotlinTreeParserVisitor extends KtVisitor<J, ExecutionContext> {
     private J.AssignmentOperation.Type mapAssignmentOperationType(KtOperationReferenceExpression operationReference) {
         IElementType elementType = operationReference.getOperationSignTokenType();
 
-        if (elementType == KtTokens.PLUSEQ)
+        if (elementType == KtTokens.PLUSEQ) {
             return J.AssignmentOperation.Type.Addition;
-        if (elementType == KtTokens.MINUSEQ)
+        } else if (elementType == KtTokens.MINUSEQ) {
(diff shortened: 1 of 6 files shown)


src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
```diff
@@ -3259,6 +3259,8 @@ private K.Binary.Type mapKBinaryType(KtOperationReferenceExpression operationRef
             return K.Binary.Type.IdentityNotEquals;
         } else if (elementType == KtTokens.ELVIS) {
             return K.Binary.Type.Elvis;
+        } else if (elementType == KtTokens.PERCEQ) {
+            return K.Binary.Type.ModAssign;
```
knutwannheden, line 3263: This should be a `J.AssignmentOperation` or similar. In Java we also have `%=`: https://github.com/openrewrite/rewrite/blob/067c32ed93aa873b27d233ca6cf6f7608da1db99/rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17ParserVisitor.java#L578-L581

knutwannheden:
All other assignment operators already appear to be supported and covered by tests in https://github.com/openrewrite/rewrite-kotlin/blob/95dcdeab77ec377524518f2df1fe7a9f70a7c3e4/src/test/java/org/openrewrite/kotlin/tree/AssignmentOperationTest.java.

See also https://kotlinlang.org/docs/reference/grammar.html#assignmentAndOperator.

src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
kunli2: good call, thanks, revised it to use `J.AssignmentOperation` instead of Binary

knutwannheden (approved):

Issue #574, opened 2023-12-21
Idempotent print issue: whitespace is lost on trailing semi-colon


traceyyoshima:
The semi-colon is detected in `maybeTrailingSemicolon`, but the deepPrefix method fails to detect the whitespace.
```kotlin
              fun test ( ) {
                  while ( true ) ;
              }
```

PR #583, opened 2024-01-02
Fix issue of whitespace lost on empty while-loop body before trailing semi-colon

kunli2:
fixes https://github.com/openrewrite/rewrite-kotlin/issues/574

This happens with an empty while-loop body only, the missing whitespace should be the prefix of the empty body which is a `J.Empty`.

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
+++ b/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
@@ -1622,7 +1622,8 @@ public class KotlinTreeParserVisitor extends KtVisitor<J, ExecutionContext> {
                 deepPrefix(expression),
                 Markers.EMPTY,
                 mapControlParentheses(requireNonNull(expression.getCondition()), data).withPrefix(prefix(expression.getLeftParenthesis())),
-                expression.getBody() == null ? JRightPadded.build(new J.Empty(randomId(), Space.EMPTY, Markers.EMPTY)) : JRightPadded.build(requireNonNull(expression.getBody()).accept(this, data).withPrefix(prefix(expression.getBody().getParent())))
+                expression.getBody() == null ? JRightPadded.build(new J.Empty(randomId(), suffix(expression.getRightParenthesis()), Markers.EMPTY)) :
+                        JRightPadded.build(requireNonNull(expression.getBody()).accept(this, data).withPrefix(prefix(expression.getBody().getParent())))
         );
     }
 
--- a/src/test/java/org/openrewrite/kotlin/tree/WhileLoopTest.java
+++ b/src/test/java/org/openrewrite/kotlin/tree/WhileLoopTest.java
@@ -82,7 +82,7 @@ class WhileLoopTest implements RewriteTest {
           kotlin(
             """
               fun test ( ) {
-                  while ( true );
+                  while ( true )   ;
               }
               """
           )


src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java
```diff
@@ -1622,7 +1622,8 @@ public J visitWhileExpression(KtWhileExpression expression, ExecutionContext dat
                 deepPrefix(expression),
                 Markers.EMPTY,
                 mapControlParentheses(requireNonNull(expression.getCondition()), data).withPrefix(prefix(expression.getLeftParenthesis())),
-                expression.getBody() == null ? JRightPadded.build(new J.Empty(randomId(), Space.EMPTY, Markers.EMPTY)) : JRightPadded.build(requireNonNull(expression.getBody()).accept(this, data).withPrefix(prefix(expression.getBody().getParent())))
+                expression.getBody() == null ? JRightPadded.build(new J.Empty(randomId(), suffix(expression.getRightParenthesis()), Markers.EMPTY)) :
```
traceyyoshima, line 1625: This looks odd, what is the reasoning for `suffix(expression.getRightParenthesis())`? The function's body, should contain a `List<JRightPadded<Statement>> statements`, which I'd expect would cover this already.

Also, it would be great if we could clean up the whitespace methods, since it's unclear why there are so many, what each method does and when to use them:

- prefix and overloads
- suffix
- infix
- endFix
- endFixAndInfix
- deepPrefix
- endFixAndSuffix
- prefixAndInfix

Ideally, we would just call `whitespace()`.

Similarly, what is the purpose for `maybeTrailingSemicolonInternal` vs `maybeTrailingSemicolon`?

kunli2: `suffix(expression.getRightParenthesis())` is to get the whitespace before the empty while-loop body.
There is an alternative way to do this by using `prefix(EmptyBody)`, because `expression.getBody()` is null for this case, so I didn't find a way to get the `EmptyBody` PSI element, I think it requires a helper method to get it, but I think it's an equivalent way to use `suffix(expression.getRightParenthesis())` and solve this.

Due to the nature of PSI space locations, I didn't find an easy way to use one method like `whitespace()` to cover all the cases.

traceyyoshima: > Due to the nature of PSI space locations, I didn't find an easy way to use one method like whitespace() to cover all the cases.

I've mentioned that PSI utilities exist several times now, so, I'm a little confused by this statement. Have not look into the utilities that are used to traverse the PSI?

kunli2: I think you didn't get the point of the PSI-based parsing or misunderstood it, PSI-based parsing is completely a different approach than cursor-based parsing ( there is not a cursor so not a `whitespace()` method or so), so I am NOT talking about the PSI utilities to handle the spaces, but the way and algorithm we used to parse whitespaces.

For example, From PSI perspective,  a space before a class declaration might appear in different places, it can be either before the a `KtClass` or be the first child of a `KtClass`, so depends on the different location, we need to call different methods accordingly.

traceyyoshima: It's true, that I hadn't spent any time in the PSI, but fundamentally it is meant to be visitable in source order. That was the main reason we spent the time to rewrite the parser.

I decided to spend a couple hours after work looking into the PSI tree to attempt to understand your position and why you feel so strongly about it. I found that it is possible to collect whitespace between elements.

This is a visitor that collects the whitespace before non-whitespace elements:

```java
        StringBuilder sb = new StringBuilder();
        KtVisitorVoid visitor = new KtVisitorVoid() {
            final Set<PsiElement> visited = Collections.newSetFromMap(new IdentityHashMap<>());
            final List<PsiElement> whitespace = new ArrayList<>();
            @Override
            public void visitElement(@NotNull PsiElement element) {
                if (!(element instanceof KtFile)) {
                    if (visited.add(element)) {
                        if (element instanceof LeafPsiElement) {
                            if (isWhitespace(element)) {
                                whitespace.add(element);
                            } else {
                                for (PsiElement psiElement : whitespace) {
                                    sb.append(psiElement.getText());
                                }

                                whitespace.clear();
                                sb.append(element.getText());
                            }
                        }
                        element.acceptChildren(this);
                        if (element.getNextSibling() != null) {
                            element.getNextSibling().accept(this);
                        }
                    }
                } else {
                    element.acceptChildren(this);
                    if (!whitespace.isEmpty()) {
                        for (PsiElement psiElement : whitespace) {
                            sb.append(psiElement.getText());
                        }
                    }
                }
            }

            private boolean isWhitespace(PsiElement element) {
                return element instanceof PsiWhiteSpace || element instanceof PsiComment;
            }
        };
```

The output from the visitor is idempotent to the text of each KtFile for all of the tests. At this point, I'm certain it's possible; I'll rewrite the parser over a weekend.

knutwannheden: The PSI can definitely be queried for all consecutive whitespace at any position in the tree and thus allow for a `whitespace()` like function to be implemented. This was basically the idea I had with `deepSpace()`, but I think it may still be incomplete.

PSI has some rules according to which it assigns whitespace as either leading or trailing whitespace of an element. While we as much as possible try to use leading whitespace in our LST models, the PSI models will typically have trailing comments (on the same line) as trailing whitespace. That is probably the biggest difference, which can always cause some problems for us.

traceyyoshima: The visitor collects the whitespace and comments between elements into a list, which can be transformed into a `Space`. But I didn't implement the transformation since it's just a quick POC.

Do you think it's fine to map the Space(s) to positions in the source to forego the PSI complexity?

Another option would be to use `PsiTreeUtil` that has utility methods to traverse forward and backward skipping whitespace and comments from a PSI element, as well as return all of the elements within the range of two PSI elements. But I suspect that will be slow.

Regardless, I'll look into `deepSpace` tomorrow.

knutwannheden: Something like your utility seems fine. Whether it uses a visitor or traverses the PSI tree using recursion and / or just loops doesn't really matter, I think. I just didn't see how you are using your visitor to collect the whitespace between two elements. AFAICT, it can be used to collect all whitespace starting at a given root node and will then collect all contained whitespace plus all whitespace "owned" by following siblings.

traceyyoshima: The `if (element instanceof LeafPsiElement)` means that it collects whitespace between tokens that aren't whitespace or comments.

Issue #569, opened 2023-12-21
Idempotent print issue: On shebang in source file


traceyyoshima:
```kotlin
#!/usr/bin/env who
class Foo
```

traceyyoshima:
The shebang exists as the first child of a file if it is present. On `G.CompilationUnit`, it is represented as a nullable String, which seems correct since it is not a comment and MUST be declared at the top of the file.

Unfortunately, whitespace is overly complex and would take time to fix and/or understand. I'm assigning this to @kunli2  since he has context to all the whitespace methods. Adding the shebang to the K.CU is simple, but it causes a idempotent print issue since the new line after the shebang is not preserved in the prefix of the class declaration.

PR #584, opened 2024-01-02
Support shebang parsing

kunli2:
fixes https://github.com/openrewrite/rewrite-kotlin/issues/569

Following the same pattern in groovy, added a nullable string to `K.CompilationUnit` for shebang.


knutwannheden (approved):
Can we add more tests?

kunli2:
> Can we add more tests?

Sure, added more tests, and the scenarios that shebang following with package declaration or imports need to be tackled a bit more, updated.

PR #586, opened 2024-01-03
Bump Kotlin to 1.9.22

radoslaw-panuszewski:
## What's changed?
* Kotlin version bumped to `1.9.22`
* usages of `rewrite-java` updated to the newest snapshot (`8.13.0-SNAPSHOT`)

## What's your motivation?
Between Kotlin `1.9.0` and `1.9.10` the signature of `FirSessionFactoryHelper.createSessionWithDependencies` method was changed. In our projects we use the newest Kotlin and `./gradlew rewriteRun` fails with `NoSuchMethodError`.

## Anything in particular you'd like reviewers to focus on?
According to the test case `AnnotationTest.arrayOfCallWithInAnnotation` the `FirArrayOfCall` was replaced with `FirArrayLiteral`. Do you think any additional handling is needed for that?

## Additional context
My local build resolved the snapshot dependency of `rewrite-java` so I fixed the compilation errors. Should I do that? I assume `rewrite-kotlin` and `rewrite-java` will be released together, right?

### Checklist
- [ ] ~~I've added unit tests to cover both positive and negative cases~~
- [ ] ~~I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)~~

### DIFF
--- a/src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
+++ b/src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
@@ -27,7 +27,6 @@ import org.jetbrains.kotlin.fir.analysis.checkers.getContainingClassSymbol
 import org.jetbrains.kotlin.fir.analysis.checkers.modality
 import org.jetbrains.kotlin.fir.analysis.checkers.toRegularClassSymbol
 import org.jetbrains.kotlin.fir.declarations.*
-import org.jetbrains.kotlin.fir.declarations.impl.FirOuterClassTypeParameterRef
 import org.jetbrains.kotlin.fir.declarations.impl.FirPrimaryConstructor
 import org.jetbrains.kotlin.fir.declarations.utils.isLocal
 import org.jetbrains.kotlin.fir.declarations.utils.isStatic
@@ -924,13 +923,10 @@ class KotlinTypeMapping(
     }
 
     private fun javaArrayType(type: JavaArrayType, signature: String): JavaType {
-        val arrayType = Array(
-            null,
-            null
-        )
+        val arrayType = Array(null, null, null)
         typeCache.put(signature, arrayType)
         val classType = type(type.componentType)
-        arrayType.unsafeSet(classType)
+        arrayType.unsafeSet(classType, null)
         return arrayType
     }
 
--- a/src/main/kotlin/org/openrewrite/kotlin/internal/PsiElementAssociations.kt
+++ b/src/main/kotlin/org/openrewrite/kotlin/internal/PsiElementAssociations.kt
@@ -242,8 +242,8 @@ class PsiElementAssociations(val typeMapping: KotlinTypeMapping, val file: FirFi
                         }
                     }
 
-                    is FirArrayOfCall -> {
-                        // `FirArrayOfCall` is not a `FirFunctionCall`, so a `JavaType$Method` is impossible.
+                    is FirArrayLiteral -> {
+                        // `FirArrayLiteral` is not a `FirFunctionCall`, so a `JavaType$Method` is impossible.
                         // The expression contains a type ref of the parameterized type, but cannot be added to a `MethodInvocation`.
                         null
                     }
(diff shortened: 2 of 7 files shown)


src/main/kotlin/org/openrewrite/kotlin/KotlinTypeMapping.kt
```diff
@@ -27,7 +27,6 @@ import org.jetbrains.kotlin.fir.analysis.checkers.getContainingClassSymbol
 import org.jetbrains.kotlin.fir.analysis.checkers.modality
 import org.jetbrains.kotlin.fir.analysis.checkers.toRegularClassSymbol
 import org.jetbrains.kotlin.fir.declarations.*
-import org.jetbrains.kotlin.fir.declarations.impl.FirOuterClassTypeParameterRef
```
radoslaw-panuszewski, line 30: This class was moved from `declarations.impl` to `declarations` so now the star (*) import catches it.

src/main/kotlin/org/openrewrite/kotlin/internal/PsiElementAssociations.kt
```diff
@@ -242,8 +242,8 @@ class PsiElementAssociations(val typeMapping: KotlinTypeMapping, val file: FirFi
                         }
                     }
 
-                    is FirArrayOfCall -> {
-                        // `FirArrayOfCall` is not a `FirFunctionCall`, so a `JavaType$Method` is impossible.
+                    is FirArrayLiteral -> {
+                        // `FirArrayLiteral` is not a `FirFunctionCall`, so a `JavaType$Method` is impossible.
```
radoslaw-panuszewski, line 246: Is this empty branch still needed? I feel like this comment is not appropriate anymore 🤔

traceyyoshima:
Hi @radoslaw-panuszewski, thank you for the PR!

I've seen a similar issue before between `1.8.0` and `1.9.0`, but haven't been able to reproduce the problem. Do you have a public project where the error occurs so that I can take a look?

My concern is that updating to `1.9.20` might cause incompatibility with `1.9.0`. If so, we'll need multiple versions of the KotlinParser instead of bumping the version up to `1.9.20`.

src/main/kotlin/org/openrewrite/kotlin/internal/PsiElementAssociations.kt
traceyyoshima: Yes, this can be remove -- FirArrayLiteral clearly communicates it's not a method invocation, thank you!

radoslaw-panuszewski:
> Hi @radoslaw-panuszewski, thank you for the PR!
> 
> I've seen a similar issue before between `1.8.0` and `1.9.0`, but haven't been able to reproduce the problem. Do you have a public project where the error occurs so that I can take a look?
> 
> My concern is that updating to `1.9.20` might cause incompatibility with `1.9.0`. If so, we'll need multiple versions of the KotlinParser instead of bumping the version up to `1.9.20`.

Hi @traceyyoshima, I realized what was the problem in my configuration. Accidentally, the Gradle build with my custom recipe applied the `spring-boot-dependencies` like this:
```kotlin
implementation(platform("org.springframework.boot:spring-boot-dependencies:3.2.0"))
```
so after adding it to the `rewrite` configuration:
```kotlin
rewrite("com.example.my-custom-recipe")
```
the platform dependency for `spring-boot-dependencies` was transitively resolved and it constrained version of the `kotlin-compiler-embeddable` to `1.9.20`. 

After removing the platform dependency, the `NoSuchMethodException` is no longer thrown. The `kotlin-stdlib` is still in version `1.9.20` (because my recipe is written in Kotlin) but the `kotlin-compiler-embeddable` is at `1.9.0` (because `rewrite-kotlin` brings it transitively). It also works in the opposite way: `rewrite-kotlin` built with Kotlin `1.9.20`  can be applied for project using Kotlin `1.9.0`.

To conclude: my problem is already resolved, but since I already opened this PR, maybe we can just merge it? 😉

src/main/kotlin/org/openrewrite/kotlin/internal/PsiElementAssociations.kt
radoslaw-panuszewski: Redundant when branch removed

radoslaw-panuszewski:
@traceyyoshima @kunli2 @knutwannheden please take a look :)

knutwannheden (approved):
LGTM. 🙏

PR #614, opened 2024-10-17
Refactor mutable field in `EqualsMethodUsage`

Laurens-W:
## What's changed?

Refactored mutable field to be scoped better

## Anyone you would like to review specifically?

@knutwannheden 

### Checklist
- [ ] I've added unit tests to cover both positive and negative cases
- [x] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)


timtebeek:
The motivation here is that it got flagged by our best practices recipes, where we do not want mutable static fields in recipes.

Following [this change](https://github.com/openrewrite/rewrite-kotlin/commit/f793f5b1ba326937091e32d64bd94a37eddc61d6) it should already no longer get flagged, so then it's merely a question of whether we want to keep this field in the recipe as a static field, or move it to every visitor instance. In the latter case it might not even make much sense anymore to store the J.Binary as a field, but I don't know enough of the motivation to cache this field and any implications on performance.

Issue #616, opened 2024-11-15
Support `.dependsOn` for KotlinParser


tschut:
## What problem are you trying to solve?
I like to have isolated and minimal tests. With the JavaParser, I can use `JavaParser::dependsOn` to load additional sources from my test. There is no equivalent in the KotlinParser.

## Have you considered any alternatives or workarounds?
Also see slack thread: https://rewriteoss.slack.com/archives/C01A843MWG5/p1731577323576189

The workaround would be to use .classpath, however, sometimes it is preferred to not add .jars just for a test.

## Are you interested in [contributing this feature to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#new-features)?

Yes, I'm willing to create a PR for this.

PR #617, opened 2024-11-15
Add .dependsOn method for KotlinParser

tschut:
## What's changed?
Add `.dependsOn` method for KotlinParser

## What's your motivation?
- https://github.com/openrewrite/rewrite-kotlin/issues/616

## Anything in particular you'd like reviewers to focus on?
* The way I have to remove the sources from the returned list is not so nice. Perhaps there is a better way without refactoring the whole thing?
* Need to add tests

## Anyone you would like to review specifically?
@timtebeek 

### Checklist
- [x] I've added unit tests to cover both positive and negative cases
- [x] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)

### DIFF
--- a/src/main/java/org/openrewrite/kotlin/KotlinParser.java
+++ b/src/main/java/org/openrewrite/kotlin/KotlinParser.java
@@ -19,7 +19,6 @@ import kotlin.Pair;
 import kotlin.Unit;
 import kotlin.annotation.AnnotationTarget;
 import kotlin.jvm.functions.Function1;
-import kotlin.random.Random;
 import lombok.AccessLevel;
 import lombok.RequiredArgsConstructor;
 import org.intellij.lang.annotations.Language;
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/kotlin/KotlinParser.java
```diff
@@ -201,7 +206,8 @@ public Stream<SourceFile> parseInputs(Iterable<Input> sources, @Nullable Path re
                                     return (SourceFile) null;
                                 })
                                 .limit(1))
-                .filter(Objects::nonNull);
+                .filter(Objects::nonNull)
+                .filter(source -> !source.getSourcePath().getFileName().toString().matches("dependsOn-[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\\.kt"));
```
timtebeek, line 210: What made you pick a generated file name? Might be nice to [similarly pick a reasonable filename](https://github.com/openrewrite/rewrite/blob/9dd7034d5b2479421ac0ccb57051d8886fac4c38/rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java#L428-L449) where we can here as well, as compared to Java.

tschut: I've copied the logic from JavaParser and polished it a bit + some small changes to make it Kotlin compatible

PR #624, opened 2025-01-11
`int` literals should return an `Integer` value

timtebeek:
## What's your motivation?
As discovered on
- https://github.com/openrewrite/rewrite-spring/pull/663

### DIFF
--- a/gradle/licenseHeader.txt
+++ b/gradle/licenseHeader.txt
@@ -1,4 +1,4 @@
-Copyright 2023 the original author or authors.
+Copyright ${year} the original author or authors.
 <p>
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


barbulescu:
@timtebeek maybe this is the reason: https://github.com/JetBrains/kotlin/blob/403acc0dee98791e7f2a068feb5b09f13c7909db/compiler/psi/src/org/jetbrains/kotlin/parsing/ParseUtils.kt#L38

timtebeek:
Thanks yes that looks to be it! Now we have to see at which point we can alter that value :)

barbulescu:
Maybe here:
```
    @Override
    public J visitConstantExpression(KtConstantExpression expression, ExecutionContext data) {
        IElementType elementType = expression.getElementType();
        Object value;
        if (elementType == KtNodeTypes.INTEGER_CONSTANT || elementType == KtNodeTypes.FLOAT_CONSTANT) {
            value = ParseUtilsKt.parseNumericLiteral(expression.getText(), elementType);
            if (value instanceof Long) {
                int intValue = ((Long) value).intValue();
                if (Objects.equals(value, (long) intValue)) {
                    value = intValue;
                }
            }
        } else if (elementType == KtNodeTypes.BOOLEAN_CONSTANT) {
```

timtebeek:
Nice, thanks! Went for a slightly different approach to ensure true longs remain longs. With that I think we're good! :)

barbulescu:
Kotlin forum is in read-only mode so I created a [bug report](https://youtrack.jetbrains.com/issue/KT-74375/parseNumericLiteral-returns-Long-when-Int-is-expected).

timtebeek:
Much appreciated! I'll have another look at your PR in rewrite-spring now that this is sorted. :)
- https://github.com/openrewrite/rewrite-spring/pull/663

PR #626, opened 2025-01-11
Method matcher should match Java primitive types in Kotlin

timtebeek:
## What's changed?
Add a test to check that we can match Java primitive types in Kotlin

## What's your motivation?
Saw that we do not match, as the method type uses `kotlin.Int`, not any Java `int`, which we then fail to match.

### DIFF
--- a/src/test/java/org/openrewrite/kotlin/MethodMatcherTest.java
+++ b/src/test/java/org/openrewrite/kotlin/MethodMatcherTest.java
@@ -58,4 +58,64 @@ class MethodMatcherTest implements RewriteTest {
           )
         );
     }
+
+    @Test
+    void matchesMethodPatternOfJavaMethodWithAnyArgument() {
+        rewriteRun(
+          spec -> spec.recipe(toRecipe(() -> new KotlinIsoVisitor<>() {
+              private static final MethodMatcher methodMatcher = new MethodMatcher("java.lang.Math max(..)");
+              @Override
+              public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext p) {
+                  if (methodMatcher.matches(method)) {
+                      return SearchResult.found(method);
+                  }
+                  return super.visitMethodInvocation(method, p);
+              }
+          })),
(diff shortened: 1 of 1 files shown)


timtebeek:
I've briefly explored doing something like the following, but it's not pretty, and not sure if such an approach will fully work:
```java
    private JavaType.@Nullable Method methodInvocationType(PsiElement psi) {
        JavaType.Method psiMethodType = psiElementAssociations.methodInvocationType(psi);

        // Look for method in the declaring type to find the correct method parameter types, not the Kotlin equivalents
        JavaType.FullyQualified declaringType = psiMethodType.getDeclaringType();
        for (JavaType.Method m : declaringType.getMethods()) {
            if (m.getName().equals(psiMethodType.getName())) {
                List<JavaType> declParameterTypes = m.getParameterTypes();
                List<JavaType> psiParameterTypes = psiMethodType.getParameterTypes();
                if (declParameterTypes.size() == psiParameterTypes.size()) {
                    for (int i = 0; i < declParameterTypes.size(); i++) {
                        JavaType declType = declParameterTypes.get(i);
                        JavaType psiType = psiParameterTypes.get(i);

                        // TODO compare types JavaType.Primitive.Int and JavaType.Class kotlin.Int, and similar friends
                    }
                }
            }
        }
        return psiMethodType;
    }
```

timtebeek:
Alternatively we can map Kotlin Primitives to JavaType.Primitive, instead of the Java.Class instances they are now; That would mean recipes written for Java that might use these primitives in their method patterns would match again. We could extend this to only do so for types coming from Java if needed / desired. Let's agree to this approach before I update the failures in
https://github.com/openrewrite/rewrite-kotlin/blob/f7417f68935829fcfe865accee4124843288216d/src/test/java/org/openrewrite/kotlin/KotlinTypeMappingTest.java#L48

timtebeek:
Any thoughts as to the `kotlin.Int` mapping to `int` here @knutwannheden ? I feel that would help recipe reuse across languages..

